Issue #122, opened 2016-06-27
Create more "smart" components?


trevorgerhardt:
The connect function isn't only meant to be used once. `scenario-editor.js` is turning into a "monster" and it might be better if we connect further down the chain.

PR #126, opened 2016-06-30
refactor(scenario-map): Connect ScenarioMap directly to redux store.

dmin:
Use react-redux's `connect` function to make `ScenarioMap` a "smart" component. Reduces coupling
between `ScenarioEditor` and `ScenarioMap`. Work in progress towards issue #122.

I didn't have any "real" data to test the changes with, but I did what I could put together some fake data to use for testing by examining your graphql query and `project-store`.

I tried to match your style and conventions as closely as I could, but if there are any issues I'd be happy to correct them.

(You might be wondering why a random person is sending in a PR - I came across Conveyal in a job listing for a javascript engineer position, and since your team has a bunch of code on github I thought I'd dig through it a little bit and get a feel for some of the projects you're working on. Interesting stuff!)

### DIFF
--- a/lib/map/scenario-map.js
+++ b/lib/map/scenario-map.js
@@ -1,6 +1,9 @@
 /** A map component showing a scenario */
 
 import React, { Component, PropTypes } from 'react'
+import {connect} from 'react-redux'
+import {setMapState} from '../actions'
+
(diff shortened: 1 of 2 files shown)


trevorgerhardt:
Thanks @dmin! Looks good!

PR #127, opened 2016-06-30
docs(README): Update mastarm usage.

dmin:
Update "how to run" instructions to use the `build --serve` command with mastarm.

This is how I was able to get scenario-editor to build/run anyway. I might just be using mastarm incorrectly...?  Came across this while working in #126.


trevorgerhardt:
Thanks @dmin!

PR #134, opened 2016-08-01
Edit modifications in second dock

catherinesnow:
https://github.com/conveyal/scenario-editor/issues/128

Was playing around with trying to get the docks to still be resizable and resize the map on edit. I couldn't figure out a good way to do this via react. Right now they are fixed sizes but still using react-dock and the map isn't resizing when the second dock is activated. If there is a good way to get the dockWidth from application.js to the children then I think it would be relatively easy enough to figure out.

Also fixed some typos and little quirks I found as I was working on this.


trevorgerhardt:
Is it possible to accomplish this without an additional action? I originally thought the implementation would use the "active" modification action and status to show the dock. Collapsing  the dock would set the active modification to null.

trevorgerhardt:
I think the standard caret / chevron style is facing right for closed and down for open. Let's switch to that style. I know I may have messed this up in the ticket and elsewhere in the application but after using it in this particular change it's more obvious to me.

catherinesnow:
@trevorgerhardt Yeah, I should have asked about this. I was under the impression that one could make  edits via the map and might want to close the extra panel. But of course I'm not as familiar with the functionality. Let me know if I should go ahead and update that.

trevorgerhardt:
Yup! Please do!

catherinesnow:
@trevorgerhardt Ok, updated! I noticed in master that deleting a modification throws an error. But, I was happy to see a bug I thought was in my branch went away when I merged in master. Whew!

I also updated with the resizing functionality for the docks and map.

trevorgerhardt:
@catherinesnow I should really be using a development branch :)

Looks great! Just merged. Appreciate the work. I'll be following up more later this week.

PR #150, opened 2016-09-27
Add ability to test with Jest

evansiroky:
Added testing functionality and a bunch of simple tests


evansiroky:
Unfortunately I'm not able to find an immediate solution to removing the `.babelrc` file.  [The Jest docs say this file is required](http://facebook.github.io/jest/docs/getting-started.html#babel-integration).  I opened up [a question on StackOverflow](http://stackoverflow.com/questions/39735459/avoid-making-babelrc-file-to-use-for-testing-with-jest) to see if anyone can come up with a solution.  Once a solution is found I'm all for adding the fix, but I'm not seeing a way at the moment.

evansiroky:
I figured it out!  See https://github.com/conveyal/mastarm/pull/36.

trevorgerhardt:
Nice! Let's get that merged in to Mastarm and then update this pull.

PR #153, opened 2016-09-28
Testing with mocking

evansiroky:
Refactor a few simple component tests so that dependent components are mocked.  This now allows development to occur on individual components without causing tests to fail on components which depend on the component being changed.


evansiroky:
Closing so I can merge into a different branch

PR #158, opened 2016-09-30
Isochrone

trevorgerhardt:
* Update isochrones on search if enabled
* Use slider to change cutoff
* Turn gray on changes to scenario or modifications

@mattwigway please review


trevorgerhardt:
@evansiroky what is the command/right way to update snapshot tests?

__tests__/components/modification/__snapshots__/group.js.snap
```diff
@@ -52,7 +52,7 @@ exports[`ModificationGroup renders correctly 1`] = `
       }
       name="Test Modification"
       projectId="1234"
-      replaceModification={[Function mockConstructor]}
+      replaceModification={[Function anonymous]}
```
evansiroky: I think these Function name differences will also be fixed in Jest 16.

PR #160, opened 2016-10-04
Add isochrone refresh button

trevorgerhardt:


trevorgerhardt:
@evansiroky Added a test for the isochrone component but it currently fails due to the `react-leaflet` portion. Any advice on how to make it work?

evansiroky:
You can and should make mocks for all sub-components that are rendered within the component you are testing.  Check out the test I made of [direction-icon.js](https://github.com/conveyal/scenario-editor/blob/2ea9625d65ab7e731b81c83671bdc4c325806d97/__tests__/components/direction-icon.js).  In this case, you wouldn't need to make a mock for leaflet since you aren't directly using it.

PR #165, opened 2016-10-10
Browsochrones

mattwigway:
Here's code integrating browsochrones for single point analysis.


lib/containers/map-control.js
```diff
@@ -8,15 +8,16 @@ import {ISOCHRONE_COMPONENT} from '../constants/map'
 function mapStateToProps (state, props) {
   const {analysis, mapState, scenario} = state
   return {
-    bundleId: scenario.bundleId,
+    bundleId: scenario.currentBundle ? scenario.currentBundle.id : null,
     center: mapState.center,
     geocoderApiKey: process.env.MAPZEN_SEARCH_KEY,
     isochroneCutoff: analysis.isochroneCutoff,
-    isochroneLatLng: analysis.isochroneLatLng,
+    // TODO duplicate code below and in containers/isochrone
```
trevorgerhardt: what is this todo for? 🤔

lib/reducers/analysis.js
```diff
@@ -16,17 +16,14 @@ export const reducers = {
   'set isochrone latlng' (state, action) {
     return {
       ...state,
-      isochroneLatLng: null
+      isochroneLatLng: action.payload
     }
   },
   'set isochrone results' (state, action) {
-    const isochrones = action.payload.avgCase.isochrones
+    let { isochrone } = action.payload
```
trevorgerhardt: Generally use `const` whenever you can. It gives us a free check that says "this variable will never be reassigned". Not immutable, but it's enough that I'd say the two extra letters are worth it.

lib/utils/browsochrones.js
```diff
@@ -0,0 +1,226 @@
...
+    scenarioHash: null,
+    origin: { x: NaN, y: NaN },
+    staticSiteRequest: null,
+    browsochrones
+  }
+}
+
+/** Attempt to fetch a file from the server, retrying on a 202 (Worker starting up) response */
+async function retryFetch (url, opts, { timeoutMs = RETRY_TIMEOUT_MILLISECONDS, authenticated = false }) {
+  let fetchFunc = authenticated ? authenticatedFetch : fetch
```
trevorgerhardt: Couldn't you always use `authenticatedFetch`? It just adds the `Authorization` header if you're logged in. But won't stop you from `fetch`ing if you're not.

https://github.com/conveyal/scenario-editor/blob/dev/lib/utils/authenticated-fetch.js#L13

lib/utils/browsochrones.js
```diff
@@ -0,0 +1,226 @@
...
+    }, {
+      authenticated: true
+    }).then(res => res.arrayBuffer())
+
+    let { browsochrones } = instance
+
+    await Promise.all([
+      browsochrones.setQuery(metadata),
+      browsochrones.setStopTrees(stopTrees),
+      browsochrones.setTransitiveNetwork(metadata.transitive) // TODO shouldn't have to do this.
```
trevorgerhardt: Is this a new way of retrieving data or have we always done it this way?

lib/utils/browsochrones.js
```diff
@@ -0,0 +1,226 @@
...
+  bikeTime: 1,
+  bikeTrafficStress: 4,
+  boardingAssumption: 'RANDOM',
+  monteCarloDraws: 220
+}
+
+// Map from scenario ID to object containing Browsochrones instance, hash of last used scenario
+// x, y coords of origin
+// TODO LRU cache
+let instances = new Map()
```
trevorgerhardt: how many browsochrone instances are we going to be holding here???

trevorgerhardt:
globals isn't technically the right word there, but it fits

lib/utils/browsochrones.js
mattwigway: It's a good point. I was thinking I also needed to use this function to grab grids from S3, but of course I don't as S3 should never need to be retried.

lib/utils/browsochrones.js
mattwigway: Possibly 3-4

trevorgerhardt: Is the idea for using an LRU to limit it to 4?

mattwigway: Yep.

PR #166, opened 2016-10-12
Test with enzyme

evansiroky:


evansiroky:
Parts of the leaflet mock are taken from react-leaflet with some extra stuff added.

PR #170, opened 2016-10-15
Add tests for scenario-map components

evansiroky:
That should be it for testing components!

### DIFF
--- a/__tests__/components/__snapshots__/frequency-entry.js.snap
+++ b/__tests__/components/__snapshots__/frequency-entry.js.snap
@@ -35,7 +35,20 @@ exports[`Component > FrequencyEntry renders correctly 1`] = `
                 ],
                 "type": "LineString",
               },
-              "stops": Array [],
+              "stops": Array [
+                Object {
+                  "stop_id": "1",
+                  "stop_lat": 36.9707,
+                  "stop_lon": -122.0246,
+                  "stop_name": "Santa Cruz",
+                },
+                Object {
+                  "stop_id": "2",
+                  "stop_lat": 37.049,
+                  "stop_lon": -122.0279,
+                  "stop_name": "Scotts Valley",
+                },
+              ],
               "trips": Array [
                 Object {
                   "duration": 1234,
@@ -82,7 +95,20 @@ exports[`Component > FrequencyEntry renders correctly 1`] = `
                 ],
                 "type": "LineString",
               },
-              "stops": Array [],
+              "stops": Array [
+                Object {
+                  "stop_id": "1",
+                  "stop_lat": 36.9707,
+                  "stop_lon": -122.0246,
+                  "stop_name": "Santa Cruz",
+                },
+                Object {
+                  "stop_id": "2",
+                  "stop_lat": 37.049,
+                  "stop_lon": -122.0279,
+                  "stop_name": "Scotts Valley",
+                },
+              ],
               "trips": Array [
                 Object {
                   "duration": 1234,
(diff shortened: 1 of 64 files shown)


__tests__/components/__snapshots__/frequency-entry.js.snap
```diff
@@ -35,7 +35,20 @@ exports[`Component > FrequencyEntry renders correctly 1`] = `
                 ],
                 "type": "LineString",
               },
-              "stops": Array [],
+              "stops": Array [
```
mattwigway: Is all of this autogenerated, or have we edited the snapshots manually? Do we need to be careful not to accidentally overwrite this?

__tests__/scenario-map/draw-polygon.js
```diff
@@ -0,0 +1,20 @@
...
+drawMock()
+
+describe('Scenario-Map > DrawPolygon', () => {
+  it('works correctly', () => {
+    // initialize
+    const mockCallback = jest.fn()
+    const drawer = new DrawPolygon(mockCallback)
+
+    // add to map
+    const mockMap = { on: jest.fn() }
```
mattwigway: I don't see any assertions here?

__tests__/scenario-map/index.js
```diff
@@ -0,0 +1,55 @@
...
+      <Provider store={mockStore}>
+        <ScenarioMap
+          {...props}
+          />
+      </Provider>
+      , {
+        attachTo: document.getElementById('test')
+      }
+    )
+    Leaflet.map // TODO test rendering of various modifications
```
mattwigway: What does this line do?

__tests__/scenario-map/stop-layer.js
```diff
@@ -0,0 +1,37 @@
...
+      <Map>
+        <StopLayer
+          {...props}
+          />
+      </Map>
+      , {
+        attachTo: document.getElementById('test')
+      }
+    )
+    expect(Leaflet.circleMarker.mock.calls[0][0]).toMatchSnapshot()
```
mattwigway: Is this asserting that the two circle markers are identical?

__tests__/components/__snapshots__/frequency-entry.js.snap
evansiroky: I added more data in the `test-utils/mock-data.js` mock objects, so this new data got percolated to all of the components that use this data.  I hope that those mock objects are stable now.

__tests__/scenario-map/index.js
evansiroky: It makes the Standard linter happy that Leaflet was used in the code.  When I import `mock-leaflet` it hijacks the original Leaflet and does some background operations that enable react-leaflet components to be mounted into JSDOM.  However, it gets tricky when I can't figure out how to make some assertions, so I hope to revisit this in the future.

trevorgerhardt: @evansiroky you can import without assigning.  Switching `import Leaflet from '../../test-utils/mock-leaflet'` to `import '../../test-utils/mock-leaflet'` would remove the standard error.

__tests__/scenario-map/stop-layer.js
evansiroky: The lines are different.  Line 33 asserts that the first circleMarker was created and line 34 asserts that the second one was also created.  You can check the [snapshot](https://github.com/conveyal/scenario-editor/blob/df253506c415b7271e9ed85b2e4f228529774881/__tests__/scenario-map/__snapshots__/stop-layer.js.snap) to see the difference.

trevorgerhardt: @evansiroky since the `Leaflet.circleMarker.mock.calls[0][0]` portion isn't clear would you mind including a comment or wrapping it in a function that describes what you are accessing and why?

PR #181, opened 2016-10-27
Fix test warnings

trevorgerhardt:
Lot's of small changes here based mainly on clearing all of the warnings shown during `npm test`:

- Refactored a few components that are still using `replaceModification` when `update` was more appropriate.
- Changed `ariaLabel` to `aria-label`. Aria props are hyphenated cc @mattwigway 
- Removed unused Title component.
- Removed the dependency on the `recompose` library by creating an internal `pure` function based on `lodash.isequal` which we were already using in the `DeepEqual` component.
- Refactored the `Input`component to be locally controlled. It always has the value in state, but can still be updated with `props`. Should probably be pulled out and shared across projects. From my experience this is the most straightforward version of how a React input should handle state and updates.


evansiroky:
The snapshots look a lot better!  Nice work!

PR #190, opened 2016-11-15
v0.1.0

trevorgerhardt:


evansiroky:
oh man, this massive list of commits makes me think we should be merging into master more often

trevorgerhardt:
@evansiroky agreed, it's a good way to know what commits are live

PR #191, opened 2016-11-15
Reduce test memory footprint

trevorgerhardt:


evansiroky:
If we're going to be modifying these snapshots with this kind of reckless abandon, then are they really providing any value?

trevorgerhardt:
Any snapshots stick out to you? I checked over most of them.

trevorgerhardt:
Also, this PR was made so that tests could actually finish on TravisCI. Compromises were obviously going to be made.

__tests__/report/__snapshots__/reroute.js.snap
```diff
@@ -1,139 +1,238 @@
...
-    </tr>
-    <tr>
-      <th>
-        Original length
-      </th>
-      <td>
-        <Distance
-          km={45.223652927700726}>
+<div>
+  <h3>
```
evansiroky: Take this snapshot for example, the largest change in this PR.  A few points:

- I do like that the h3 tag is now in the snapshot since the original way tested just the table and map, but not the h3.
- We're now flip-flopping in how we test leaflet rendering back to mocking React-Leaflet.  This is what I did initially, but it is prone to reporting false test passes for when Leaflet changes and should cause a failure.  However, the other way is sort of equally as bad if not worse since we're relying on private method mocking of Leaflet, so it could report false test passes or failures for when it should fail or work respectively.
- And when you say `I checked over most of them` I begin to question what value these snapshots are giving us if some don't need to be checked.  I've been guilty too of skimming over snapshots or even not understanding the whole thing and just accepting the output.  Worse still I have also been guilty of checking in snapshots that are totally wrong.  My point here is that it doesn't seem like we're really asserting something specific enough that we can conclusively say we know what we're actually testing.  Perhaps if we do continue to use snapshots we should impose a limit on the length of the snapshot file so we can easily understand exactly what we're asserting.

__mocks__/react-leaflet.js
```diff
@@ -0,0 +1,20 @@
+/* global jest */
+
+import React from 'react'
+
+const ReactLeaflet = require.requireActual('react-leaflet')
+
+// mock tile layer urls
+process.env.LEAFLET_TILE_URL = 'mock.url/tile'
+
+ReactLeaflet.GeoJson.prototype.render = function () {
```
evansiroky: I guess I have to take back my comment on how we're flip-flopping on mocking React-Leaflet.  Instead with this commit we're doubling down on mocking private methods of both Leaflet and now also React-Leaflet.  If React-Leaflet ever decides to modify the render method of Marker or GeoJson, our tests won't replicate the actual situation.  And the purpose of this is to include the rendering of the marker in the snapshot despite that the actual html will not look like this anyways?  At least with the previous testing method we were verifying that Leaflet was receiving a call to draw a marker which is the actual trigger of drawing the marker in the browser.

PR #192, opened 2016-11-15
Spectrogram configurable colors

mattwigway:
Allow configuring the colors in the spectrogram, and fix a sorting bug in the regional analysis.


mattwigway:
@trevorgerhardt can you review?

PR #193, opened 2016-11-17
Rearrange

trevorgerhardt:
Big rearrangement of files that 

* Allows for shorter import paths in tests
* Puts all components in the components directory


evansiroky:
My only complaint with this is that the testing utility file `mock-data.js` is now treated as project code and is included in the [coverage reports](https://codecov.io/gh/conveyal/scenario-editor/compare/c27341111e0bbbb7756fc95762a528012bf776c2...35fa8342bd63a33387182259b516672a1ea963e5/tree?path=lib%2Futils).  Try to move it into a mocks directory or include an argument to mastarm with `--test-path-ignore-patterns` to ignore it.

PR #194, opened 2016-11-28
Refactor browsochrones

trevorgerhardt:


index.html
```diff
@@ -9,6 +9,6 @@
   </head>
   <body>
     <div id="root"></div>
-    <script src="/assets/index.js"></script>
+    <script src="/assets/index.js" async></script>
```
mattwigway: What does this do, out of curiosity?

lib/utils/browsochrones.js
```diff
@@ -3,15 +3,16 @@
...
 import convertToR5Modification from './convert-to-r5-modification'
 import authenticatedFetch from './authenticated-fetch'
-import timeout from './timeout'
 
-/** How often to retry 202 response from the server */
-const RETRY_TIMEOUT_MILLISECONDS = 10 * 1000
+const debug = dbg('scenario-editor:browsochrones')
+
+function assert (e, msg) {
+  if (!e) throw new Error(msg)
```
mattwigway: The parameter name `e` is not particularly descriptive, and in fact would suggest to me that this is a node-style callback with error as first arg.

lib/utils/browsochrones.js
```diff
@@ -47,21 +47,21 @@ let instances = new Map()
...
 
 /** get a new Browsochrones instance with grids already loaded */
 async function newInstance () {
-  if (grids == null) throw new Error('Grids are not loaded!')
+  if (!grids) throw new Error('Grids are not loaded!')
 
   let browsochrones = new Browsochrones()
 
-  await grids.map(({ grid, name }) => browsochrones.putGrid(name, grid))
+  await Promise.all(grids.map(({ grid, name }) => browsochrones.putGrid(name, grid)))
```
mattwigway: Ah, that's why it kept breaking.

index.html
trevorgerhardt: Script tags without `async` or `defer` "delays the first paint". I did an audit of the scenario editor with Google's [Lighthouse](https://github.com/GoogleChrome/lighthouse) and that was the easiest/only change that I've made so far. More to come (probably)

trevorgerhardt:
@mattwigway changed it from `e` to `value`

PR #199, opened 2016-11-30
feat(analysis): Lazy-load grids, rather than loading them all up front.

mattwigway:
Grids used to be loaded on project load. Since most were never used, this was a huge waste of memory

in larger projects.


mattwigway:
This lazy loads grids when requested rather than loading them all at the start and potentially crashing the browser in large cities. It also switches to the dot-density representation for opportunities.

mattwigway:
Despite any snide remarks from Travis, the tests do in fact pass.

lib/actions/project.js
```diff
@@ -50,8 +49,6 @@ export const load = (id) =>
     next: async (response) => {
       const project = await response.json()
       clearBrowsochrones()
```
trevorgerhardt: This is definitely a side affect we want to put into an action. 👌

trevorgerhardt: Once we get all the browsochrones stuff into the store that is.

lib/components/analysis/index.js
```diff
@@ -68,18 +65,10 @@ export default class SinglePointAnalysis extends Component {
   }
 
   async componentDidMount () {
```
trevorgerhardt: No longer needs to be `async`right?

lib/containers/opportunity.js
```diff
@@ -4,20 +4,15 @@
...
 
 function mapStateToProps ({ analysis, project }) {
   return {
-    // TODO don't duplicate grid selection logic here and in SinglePointAnalysis
-    grid: getGrids().get(analysis.currentIndicator || project.currentProject.indicators[0].key),
-    breaks: Choropleth.quantile({ noDataValue: 0 }),
-    colors: ['#feebe2', '#fbb4b9', '#f768a1', '#c51b8a', '#7a0177'],
-    labels: 16
+    grid: analysis.grid,
+    color: '#c49b7a'
```
trevorgerhardt: Put this color in the `constants` folder somewhere for now? We probably shouldn't have hex colors all over the place.

lib/actions/analysis/index.js
```diff
@@ -15,20 +15,24 @@ export const setActiveVariant = createAction('set active variant')
...
-    getIsochronesAndAccessibility({ scenarioId, bundleId, modifications, isochroneCutoff, origin, indicator, workerVersion })
-      .then(({ isochrone, accessibility, indicator, isochroneCutoff, spectrogramData }) => [
-        setIsochroneResults({ isochrone, accessibility, indicator, isochroneCutoff, spectrogramData }),
-        setIsochroneFetchStatus(false)
+    getIsochronesAndAccessibility({ scenarioId, projectId, bundleId, modifications, isochroneCutoff, origin, indicator, workerVersion })
+      .then(({ isochrone, grid, accessibility, indicator, isochroneCutoff, spectrogramData }) => [
+        setIsochroneResults({ isochrone, grid, accessibility, indicator, isochroneCutoff, spectrogramData }),
+        setIsochroneFetchStatus(false),
+        // TODO there's got to be a more elegant way to call an action after isochrone results are updated
+        next
```
trevorgerhardt: Could just pass a "nextActions" array and spread them in here. `[ setIsochroneResults(), setIsochronesFetchStatus(), ...nextActions ]`.

This whole thing will change a bit once we move them into actions and use `woonerf/fetch`. Don't worry too much about it for now.

lib/utils/browsochrones.js
```diff
@@ -65,15 +62,24 @@ async function createInstanceAndPutGrids (grids) {
...
+async function ensureGridLoaded (projectId, indicator) {
+  if (!processedGrids.has(indicator)) {
+    const grid = await authenticatedFetch(`/api/grid/${projectId}/${indicator}`)
+      .then(res => res.arrayBuffer())
+
+    // put it in all instances
+    const promises = []
+    // Map does not have .map or .forEach
+    for (const [, instance] of instances) {
+      promises.push(instance.browsochrones.putGrid(indicator, grid))
```
trevorgerhardt: You can await directly here.

```js
for (const [, instance] of instances) {
  await instance.browsochrones.putGrid(indicator, grid)
}
```

I would use `Promise.all` in this situation if you there was a possibility of a significant speed up from doing them asynchronously, like fetching external data.

lib/containers/opportunity.js
mattwigway: Do we want it in constants, or configurations?

lib/utils/browsochrones.js
mattwigway: That was why I had used promise.all, there could be quite a few instances. It may not be a significant slowdown though, although each one has to do some processing on the grid.

trevorgerhardt: Since there was confusion before just wanted to make sure you knew it was possible 👌

lib/containers/opportunity.js
trevorgerhardt: Ah, depends. We should put it in constants until we feel the need to make it dynamic. Then we should put it in the `configurations/store.yml`

PR #200, opened 2016-12-01
Broken Modification Editing Fixes

trevorgerhardt:
fix(modifications): Correct usages of _update and use an array and ch…eck manually instead of using a

### DIFF
--- a/lib/actions/get-feeds-routes-and-stops.js
+++ b/lib/actions/get-feeds-routes-and-stops.js
@@ -133,7 +133,7 @@ function routeSorter (r0, r1) {
 
 function getUniqueRouteIdsFromModifications (modifications) { // TODO: this can be done at the reducer
   // for every route referenced in the scenario we get the patterns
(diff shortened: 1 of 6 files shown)


lib/actions/get-feeds-routes-and-stops.js
```diff
@@ -143,10 +143,16 @@ function getUniqueRouteIdsFromModifications (modifications) { // TODO: this can
...
+      if (modification.routes !== null) {
+        modification.routes.forEach((r) => {
+          if (routeIds.indexOf(r) === -1) {
+            routeIds.push(r)
+          }
+        })
+      }
     }
   }
-  return [...routes]
```
mattwigway: Argh I think this is a bug in ES6... blech. Good fix.

trevorgerhardt: Except, we haven't updated mastarm and therefore babel in awhile...

PR #201, opened 2016-12-02
Comparing two scenarios

trevorgerhardt:


lib/components/analysis/histogram.js
```diff
@@ -3,7 +3,8 @@
  */
 
 import React, { Component, PropTypes } from 'react'
-import {Bar as BarChart} from 'react-chartjs'
+import chroma from 'chroma-js'
```
trevorgerhardt: We use [color](https://www.npmjs.com/package/color) in other libraries. Ideally we'd consolidate which color manipulation libraries we use. Any specific reason for choosing chroma or needing both?

trevorgerhardt:
@mattwigway open PR's as soon as you can to start the discussion and review process

lib/components/analysis/histogram.js
trevorgerhardt: Or can we just use [d3-color](https://github.com/d3/d3-color) since it's likely we'll use d3 (as we already do in Transitive).

trevorgerhardt: ?

lib/reducers/analysis.js
```diff
@@ -55,6 +58,33 @@ export const reducers = {
       activeVariant: action.payload
     }
   },
+  'set doing comparison' (state, action) {
```
trevorgerhardt: Is there better terminology we could use here? 'set doing comparison' and `doingComparison` sound weird.

lib/utils/browsochrones.js
```diff
@@ -192,7 +192,7 @@ async function ensureOriginLoadedForInstance ({
     }, true).then(res => res.arrayBuffer())
 
     await instance.browsochrones.setOrigin(origin, { x, y })
-    await instance.browsochrones.generateSurface()
+    await instance.browsochrones.generateSurface('AVERAGE')
```
trevorgerhardt: Is this the grid name? Is it an internal Browsochrones value? If it's the former, let's use an options object `generateSurface({gridName: 'AVERAGE'})` if it's the latter let's use a browsochrones exported const `browsochrones.generateSurface(browsochrones.AVERAGE)` with a default value.

trevorgerhardt:
v0.11.0 is out, no need to install from a branch: https://github.com/conveyal/browsochrones/releases

lib/reducers/analysis.js
mattwigway: I thought about isComparison but it's not a question of is or isn't. We could just use `comparison` but I think it's not as clear, or something like `comparisonActive`

trevorgerhardt: `activeComparison`?

mattwigway: I think that would be confusing, sounds like you're saying which comparison is active rather than it being a boolean.

trevorgerhardt: Oh oh oh, I lost track of what this represented. `comparisonInProgress`?

mattwigway: That could work

lib/reducers/analysis.js
```diff
@@ -61,7 +61,7 @@ export const reducers = {
   'set doing comparison' (state, action) {
```
trevorgerhardt: Reducer name also?

mattwigway:
Agh, the comment got put in the scope field on that last commit

mattwigway:
Rebased.

mattwigway:
It looks like dropping in `d3-color` will be nontrivial due to the use of color scales in the spectrogram; I'm going to hold on that and will make a separate PR.

mattwigway:
There are still some tests to write here but in the interests of avoiding divergence I'd propose we merge following a final review and I'll follow up with a testing PR. This branch is feature-complete at this point. @trevorgerhardt github won't let me request a review from you, not sure why that is.

trevorgerhardt:
Agreed. Let's merge! 👍

mattwigway:
@evansiroky can you take a gander at the tests that do exist here and if you give me a thumbs-up I'll proceed to merge?

package.json
```diff
@@ -26,7 +26,7 @@
   },
   "homepage": "https://github.com/conveyal/scenario-editor#readme",
   "dependencies": {
-    "@conveyal/gridualizer": "^1.4.0",
+    "@conveyal/gridualizer": "^1.6.0",
```
evansiroky: Might be good to bump this up to 1.7.1 to be consistent with the dependencies of npm and yarn.

PR #202, opened 2016-12-02
Fix tests and also implement testing changing a modification name

trevorgerhardt:
@mattwigway this is a good example of testing functionality you were referring too. Testing components with snapshots just makes sure there aren't obvious errors in the code. Testing a container this way you can snapshot the actual actions that occur after an event. 

Jest + Redux + Enzyme 👍 🌮 🚀 

It's kinda freakin awesome actually :)


trevorgerhardt:
Now we just need to get Travis CI to run our freakin tests...

lib/containers/__tests__/__snapshots__/modification-editor.js.snap
```diff
@@ -0,0 +1,53 @@
+exports[`Container > Modification Editor renders correctly 1`] = `
+Array [
+  Object {
+    "payload": Object {
+      "bidirectional": false,
+      "entries": Array [],
+      "feed": "1",
+      "id": "1234",
+      "name": "New Moficiation Name",
```
mattwigway: I'm not sure what a moficiation is, where is this coming from?

lib/containers/__tests__/modification-editor.js
```diff
@@ -26,10 +26,9 @@ describe('Container > Modification Editor', () => {
     )
 
     wrapper
-      .find('.InnerDock > .panel-body > .form-group > input[name="Name"]')
-      .simulate('click')
-      .simulate('keypress', 'a')
+      .find('input[name="Modification Name"]')
+      .simulate('change', {target: {value: 'New Moficiation Name'}})
```
mattwigway: Ah. Moficiation should be Modification. Or maybe Moficiation is better as it's sure not to be the default.

PR #205, opened 2016-12-12
Update Auth0 Lock look and feel

trevorgerhardt:


lib/utils/auth0.js
```diff
@@ -16,8 +17,16 @@ export const lock = authIsRequired
       },
       redirect: false
     },
-    closeable: false,
-    autoclose: true
+    allowSignUp: false,
+    theme: {
+      logo: 'http://conveyal.com/img/logo-128x128.png',
```
mattwigway: We need to serve this via SSL to avoid mixed content warnings. Let's use https://s3-eu-west-1.amazonaws.com/analyst-logos/conveyal-128x128.png

PR #208, opened 2016-12-16
feat(analysis): Allow uploading shapefile grids using multiple select.

mattwigway:


mattwigway:
@evansiroky can you take a look at the tests here?

lib/components/__tests__/__snapshots__/create-grid.js.snap
```diff
@@ -0,0 +1,301 @@
...
+                        108,
+                        117,
+                        101,
+                      ],
+                      "type": "Buffer",
+                    },
+                    Symbol(type): "",
+                    Symbol(closed): false,
+                    Symbol(name): "TEST.CSV",
+                    Symbol(lastModified): 1481916245840,
```
evansiroky: this number will change each snapshot.  Try to find a way to make the value fixed.

lib/components/__tests__/create-grid.js
```diff
@@ -0,0 +1,34 @@
...
+      finish={jest.fn()}
+      />
+
+    const csvFile = new File(['X,Y,value'], 'TEST.CSV')
+
+    const tree = mount(cg)
+
+    tree.find('input[type="file"]').simulate('change', { target: { files: [csvFile] } })
+
+    expect(mountToJson(tree)).toMatchSnapshot()
```
evansiroky: Since you are testing that a certain part of the component should change, try to find just that part of the component that changes and snapshot that area.

mattwigway: I'm testing something under the top level, so I don't know that we actually want to do that. I guess we could not do a snapshot test and just test that it exists, but the snapshot kills two birds with one stone.

lib/components/create-grid.js
```diff
@@ -53,8 +54,16 @@ export default class CreateGrid extends Component {
     finish(projectId)
   }
 
+  setFiles = (e) => {
+    const files = e.target.files
+    console.log(`setting files ${files}`)
+    console.dir(files)
+    const csv = files.length === 1 && files[0].name.toLowerCase().endsWith('.csv')
+    this.setState({...this.state, files, csv})
```
evansiroky: This will still set the state and show files if they are not of csv type.  Is that desired?

lib/components/__tests__/create-grid.js
evansiroky: Perhaps do something like this:

```javascript
expect(mountToJson(tree.find('span')).toMatchSnapshot()
```

That might find other spans tho, so you could add a classname and search for that.

lib/components/create-grid.js
mattwigway: Not sure I understand the comment. My understanding of this code is that it's putting the currently selected files in the state, as well as flag that says whether the currently selected file is a CSV file or not (it's also possible to upload a shapefile, in which case we don't need to show any additional fields).

evansiroky: ah ok, so multiple file types are allowed.  Ok, sounds good.

evansiroky:
Don't forget to fix the linting issue tho.

mattwigway:
Good call. I think lonlat is too specific (trying to only handle the names we encounter on the web in e.g. leaflet), I think we want to be more flexible and handle more possible names, but the idea is well taken. It's a little tricky as we'll have to use a FileReader to parse the first line of a (potentially very large) file. Certainly not impossible but it's a post-1.0 feature given the tight timeline there. https://app.asana.com/0/150048889193026/232402333785327

trevorgerhardt:
@mattwigway agree with @evansiroky tho. If we can add one or two more parsing checks to get 99% of the use cases then that would be a great way to do it. As it is I imagine it would currently get us > 80% of the way there.

mattwigway:
@evansiroky / @trevorgerhardt should be better now.

PR #210, opened 2016-12-19
Speed painting & Setting total duration

trevorgerhardt:
Mainly timetable changes. You can now set the total duration which automatically updates the average speed (and vice versa).

Timetables now have a `speeds` array which corresponds to the number of segments of the pattern. If you edit a segment speed, it updates the average speed and duration. 

*NOTE*: This currently will error on changes to segments out as we need to update analyst to take speeds. So we need to confirm this is the way we want to pass segment speeds to analyst and make those corresponding changes.


mattwigway:
I made the backend changes in the `speed-variation` branch of the Analyst backend. However, after loading an existing modification, the speeds are not populated.

FYI, if you want to work with your existing mongo db, there's a script in the migrations/ folder of Analyst to convert the individual speed values to an array.

trevorgerhardt:
@mattwigway so my implementation was to keep the `speed` field on a timetable and add a `speeds` field on the timetable that is an array. If you set the segment `speeds`, that array is always the same length as the number of segments on the pattern, then it calculates the average speed automatically.

But a key part is that it's on the timetable for Add Trips. From your explanation it sounded like different speeds were wanted at the timetable level, not on the modification itself. This definitely seems like it's the desired approach for `add-trip-pattern`. 

I didn't add it yet, but for reroute we can add the `speeds` field also. 

Regarding field names, is it wise to use the same field for "average speed" as for the specific feeds of each segment or should we keep them separate? Use more verbose names like `averageSpeed` and `segmentSpeeds`?

mattwigway:
Yeah, we should use the more verbose names, although I'm not sure we
actually want to include both average speed and segment speeds. I was
thinking we'd just only use segment speeds and calculate average speed on
the fly.

trevorgerhardt:
@mattwigway so just use the average speed as a UI thing to populate the `segmentSpeeds` array?

mattwigway:
Yes, that's how I'd do it.

trevorgerhardt:
@mattwigway ready to be reviewed! Failing tests are not from these changes, they are currently failing in `dev`

lib/components/modification/add-trip-pattern.js
```diff
@@ -19,13 +19,52 @@ export default class AddTripPattern extends DeepEqual {
...
+    }
+
+    this.setState({
+      ...this.state,
+      ...getStateFromProps(newProps)
+    })
+  }
+
+  _setTimetableSegmentLengthsTo (newLength) {
+    console.log('setting timetable lengths to...', newLength)
```
mattwigway: I know you wouldn't let me use console.log

lib/components/modification/segment-speeds.js
```diff
@@ -0,0 +1,57 @@
...
+  render () {
+    const {segmentSpeeds} = this.props
+    const {collapsed} = this.state
+
+    return (
+      <div>
+        <div
+          onClick={this._toggleCollapsed}
+          style={{cursor: 'pointer', paddingBottom: '12px'}}>
+          <Icon type={collapsed ? 'chevron-right' : 'chevron-down'} /> <strong>Set Individual Segment Speeds</strong>
```
mattwigway: I know it didn't exist when you created this component, but we could use the Collapsible component here.

mattwigway:
OK, fixed the getHopTimes test. The other three failing tests are fixed by #212.

lib/components/modification/add-trip-pattern.js
trevorgerhardt: 😬

lib/components/__tests__/__snapshots__/collapsible.js.snap
```diff
@@ -4,18 +4,26 @@ exports[`Component > Collapsible collapses correctly 1`] = `
...
+        role="button"
+        style={
+          Object {
+            "cursor": "pointer",
+            "paddingBottom": "12px",
+          }
+        }>
         <Icon
-          type="caret-right">
+          type="chevron-right">
```
mattwigway: Is there a reason to use a chevron instead of a caret? I feel like a caret is more standard for this type of UI element.

lib/components/modification/add-trip-pattern.js
```diff
@@ -167,6 +147,7 @@ export default class AddTripPattern extends DeepEqual {
             value={this.state.spacing}
             label={messages.transitEditor.stopSpacingMeters}
             onChange={this.onStopSpacingChange}
+            units='meters'
```
mattwigway: This should be a constant from messages to allow translation.

lib/components/modification/convert-to-frequency.js
```diff
@@ -32,10 +32,13 @@ export default class ConvertToFrequency extends DeepEqualComponent {
     })
   }
 
-  replaceEntry = (index, entry) => {
+  replaceEntry = (index, newEntryProps) => {
     const {modification, update} = this.props
     const entries = [...modification.entries]
-    entries[index] = entry
+    entries[index] = {
```
mattwigway: Why did we change this?

lib/components/modification/segment-speeds.js
```diff
@@ -0,0 +1,209 @@
...
+  _setSegmentSpeedsLengthTo (newLength) {
+    const {segmentSpeeds, update} = this.props
+    const newSegmentSpeeds = segmentSpeeds.slice(0, newLength)
+    while (newSegmentSpeeds.length < newLength) newSegmentSpeeds.push(segmentSpeeds[0] || DEFAULT_SEGMENT_SPEED)
+    update({segmentSpeeds: newSegmentSpeeds})
+  }
+
+  _confirmedSetAllSpeedsOnce () {
+    if (!this.__confirmedSetAllSpeedsOnce) {
+      this.__confirmedSetAllSpeedsOnce = window.confirm(`Setting the average speed will change all of the segment speeds. Do you wish to continue?`)
```
mattwigway: Double underscores?

lib/components/modification/segment-speeds.js
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  componentDidMount () {
+    const {segmentDistances, segmentSpeeds} = this.props
+    if (segmentDistances.length !== segmentSpeeds.length) {
+      this._setSegmentSpeedsLengthTo(segmentDistances.length)
+    }
+  }
+
+  _setSegmentSpeedsLengthTo (newLength) {
```
mattwigway: Is underscore-prefixing function names really necessary? In the case of react components generally you shouldn't call functions on them externally anyhow.

lib/components/modification/segment-speeds.js
```diff
@@ -0,0 +1,209 @@
...
+          ))}
+        </Collapsible>
+      </div>
+    )
+  }
+}
+
+function getAverageSpeedOfSegments ({segmentDistances, speeds}) {
+  const totalDistance = sum(segmentDistances)
+  const weightedSpeeds = speeds.map((s, i) => (s * segmentDistances[i]))
```
mattwigway: Nice, weighted average.

lib/components/modification/segment-speeds.js
```diff
@@ -0,0 +1,209 @@
...
+          onChange={this._setDwell}
+          units='seconds'
+          value={dwellTime}
+          />
+        <InputNumber
+          label='Average segment speed'
+          min={1}
+          onChange={this._setAllSegmentSpeedsTo}
+          units='km/h'
+          value={toPrecision(averageSpeed)}
```
mattwigway: We might want to avoid use of toPrecision here and just add a format parameter to InputNumber that takes a `d3-format` format string.

lib/components/report/__tests__/__snapshots__/add-trips.js.snap
```diff
@@ -142,7 +142,7 @@ exports[`Report > AddTrips renders correctly 1`] = `
         </td>
         <td>
           <span>
-            40 km/h (24.9 mph)
+            NaN km/h (NaN mph)
```
mattwigway: Is this intentional?

lib/components/scenario-map/stop-select-polygon.js
```diff
@@ -27,7 +27,7 @@ export default class StopSelectPolygon extends MapControl {
 
     const mod = Object.assign({}, modification)
 
-    if (action === 'add') mod.stops = [...new Set([...mod.stops, ...stops])]
+    if (action === 'add') mod.stops = [...(new Set([...mod.stops, ...stops]))]
```
mattwigway: Did set-spreading get fixed in Babel?

lib/utils/math.js
```diff
@@ -0,0 +1,13 @@
+
+/**
+ * `toFixed` leaves `0`'s at the end and converts to a String.
+ *
+ * @param {Number} n
+ * @param {Number} precision
+ * @return {Number}
+ */
+
+export function toPrecision (n, precision = 2) {
```
mattwigway: Can't we use d3-format? Is there a reason why we want it to stay a number? The general reason for using strings is that some fairly straightforward numbers cannot be represented as floating-point numbers, and the html5 number input type uses strings internally.

lib/components/modification/add-trip-pattern.js
trevorgerhardt: Yes, I've been bad about using messages due to the restart required. Wanted to address that.

lib/components/__tests__/__snapshots__/collapsible.js.snap
trevorgerhardt: Consistency across the UI. We used chevrons already for this purpose. Could make the switch everywhere?

lib/components/modification/convert-to-frequency.js
trevorgerhardt: So that you can pass just the props that changed. It's similar to using the `update` function instead of `replaceModification` function [here](https://github.com/conveyal/scenario-editor/blob/dev/lib/components/modification/editor.js#L78). For example: if you're only changing the `speed` property you could call `replaceEntry(index, {speed: newSpeed})`.

lib/components/modification/segment-speeds.js
trevorgerhardt: True. For react components it's not as relevant. Convention I picked up elsewhere that is the response to the lack of "private" functions.

For me it helps show functions that are clearly acting on the internal state of the instance.

lib/components/modification/segment-speeds.js
trevorgerhardt: Variable that is only relevant to the private method. Another convention.

lib/components/modification/segment-speeds.js
trevorgerhardt: You ever have those moments where you're just google searching the wrong term so nothing relevant comes up so you just write it yourself? Forgot about d3-format 👌

lib/components/report/__tests__/__snapshots__/add-trips.js.snap
trevorgerhardt: Hmmm, not necessarily. Good catch. I'll look into it.

lib/components/scenario-map/stop-select-polygon.js
trevorgerhardt: It is only broken in `loose` mode and I turned off loose mode.

PR #211, opened 2016-12-19
Misc fixes

mattwigway:
A few miscellaneous fixes

- Link the spectrogram scales so they are comparable when toggling between two comparison spectrograms.
- Clear all analysis data when leaving analysis mode (no extra isochrones on the map when coming back into analysis mode, also saves memory).
- Add legends to regional results.


mattwigway:
Added a few more changes, mostly to make the spectrogram look nice and not crash.

PR #212, opened 2017-01-02
New regional analysis display

mattwigway:
More additions to the new regional analysis interface. It needs a few things still but wanted to get this up there so that review can be done on what's here so far.

Still needed:
1) Automated tests, will look at tomorrow
2) Entering the regional analysis mode is currently causing a page refresh, need to fix that
3) Styling is needed, there are some spurious scrollbars and missing padding.


mattwigway:
This should be ready to go.

lib/components/analysis/__tests__/regional.js
```diff
@@ -1,100 +1,155 @@
 /* global describe, expect, it, jest */
 
+import uuid from 'uuid'
 import React from 'react'
-import renderer from 'react-test-renderer'
+import { mount } from 'enzyme'
+import { mountToJson } from 'enzyme-to-json'
 import Regional from '../regional'
 
+const mockProfileRequest = {
```
trevorgerhardt: I'm not sure if having one giant `mock-data` file is the way to go forever, but I do like the idea of pulling it out of the test itself for reuse in multiple tests. I'd move these to `utils/mock-data.js` for now and we can consider organization later.

lib/components/analysis/profile-request-display.js
```diff
@@ -0,0 +1,97 @@
...
+    request: PropTypes.shape({
+      date: PropTypes.string.isRequired,
+      accessModes: PropTypes.string.isRequired,
+      fromTime: PropTypes.number.isRequired,
+      toTime: PropTypes.number.isRequired
+    })
+  }
+
+  /** Render a mode set */
+  renderModes (modes) {
```
trevorgerhardt: I know I advocated for splitting up rendering this way but I'm now very much in favor of extracting a stateless functional component when no `this` is required.

```js
function ModeSet({modes}) {
```

and be used like

```js
<ModeSet modes={request.accessModes} />
```

lib/components/analysis/regional.js
```diff
@@ -26,23 +31,129 @@ export default class RegionalAnalysis extends Component {
...
     if (analysis == null) {
       loadRegionalAnalyses(projectId)
-      setActiveRegionalAnalysis({ id: analysisId, percentile: 'mean' })
     }
 
+    setActiveRegionalAnalysis({ id: analysisId, percentile: 'mean' })
     addRegionalAnalysisLayerToMap()
   }
 
+  selectComparisonAnalysis = chosen => {
```
trevorgerhardt: Definitely prefer parentheses for readability of fat arrow function parameters.

lib/components/analysis/regional.js
```diff
@@ -26,23 +31,129 @@ export default class RegionalAnalysis extends Component {
...
+    const {
+      setActiveRegionalAnalysis,
+      analysisId,
+      addRegionalAnalysisLayerToMap,
+      removeRegionalAnalysisLayerFromMap,
+      addRegionalComparisonLayerToMap,
+      removeRegionalComparisonLayerFromMap
+    } = this.props
+
+    if (chosen != null) {
```
trevorgerhardt: Is there a more explanatory variable name that we can use instead of `chosen` here?

lib/components/analysis/regional.js
```diff
@@ -67,6 +193,33 @@ export default class RegionalAnalysis extends Component {
...
   }
+
+  componentWillUnmount () {
+    const { removeRegionalAnalysisLayerFromMap, removeRegionalComparisonLayerFromMap } = this.props
+    removeRegionalAnalysisLayerFromMap()
+    removeRegionalComparisonLayerFromMap()
+  }
+}
+
+/**
```
trevorgerhardt: This should be pulled out into an action.

trevorgerhardt: At some point soon we're going to remove "access" to authenticated fetch so it's not so tempting to use outside of actions 😉

mattwigway:
Rebased and merged.

PR #213, opened 2017-01-04
Error reporting

mattwigway:
Display errors in the UI when there is a backend error.


mattwigway:
(should be ready to go)

PR #214, opened 2017-01-05
Exact times

mattwigway:
Add ability to create routes that have exact times (i.e. they depart at exactly the given frequency from startTime inclusive to endTime exclusive, with no variation/Monte Carlo simulation). Should be ready to go.


mattwigway:
Ack, wrong branch.

PR #215, opened 2017-01-05
Exact times

mattwigway:
Add ability to create routes that have exact times (i.e. they depart at exactly the given frequency from startTime inclusive to endTime exclusive, with no variation/Monte Carlo simulation). Should be ready to go.


mattwigway:
@trevorgerhardt make sure you take a look at the changes in `editor.js`, after merging your changes in #210 I was unable to create modifications other than add-trips and reroute, I believe I fixed the issue (it was expecting to find a segments array on all modification types).

PR #223, opened 2017-01-06
fix(browsochrones): Don't refetch origin when changing isochrone cutoff.

mattwigway:
Previously we did not keep the cached profile request up to date as parameters changed, which meant

that, once any parameter that caused refetch had been changed, we were always refetching the origin

if any parameters changed, even if they were parameters that should not cause a refetch. e.g. you

could change the date and that would correctly cause a refetch, but after that changing the

isochrone cutoff would cause a refetch, which it shouldn't.


mattwigway:
Again, merging in advance of review, but let me know if anything looks off @trevorgerhardt!

PR #229, opened 2017-01-19
Fix auth0 logout

trevorgerhardt:
This will hopefully fix the ambiguous log out error we're having on production. 😕

### DIFF
--- a/lib/reducers/map-state.js
+++ b/lib/reducers/map-state.js
@@ -9,7 +9,7 @@ export const reducers = {
     }
   },
   'log out' (state, action) {
-    return { state: null }
+    return createInitialState()
   },
   'remove component from map' (state, action) {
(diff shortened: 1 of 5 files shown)


lib/reducers/map-state.js
```diff
@@ -9,7 +9,7 @@ export const reducers = {
     }
   },
   'log out' (state, action) {
-    return { state: null }
+    return createInitialState()
```
mattwigway: Why does this change mapState?

trevorgerhardt: Application requires "center" so this was causing a React propType error on logout.

trevorgerhardt: The Application component I mean.

mattwigway: Gotcha

PR #230, opened 2017-01-19
feat(regional): Allow deleting and canceling regional analyses.

mattwigway:


mattwigway:
Should be ready to go.

PR #241, opened 2017-01-25
Fix bug that did not allow setting of times in a timetable

### LINKED ISSUE #231: Start time of timetables gets reset to 7AM
Report from Marco at Movares that when you switch to a different modification the start time in timetables always gets reset to 7AM.

Steps to reproduce:
- Create a new scenario
- Create a new "add trips" modification in that scenario
- Create a new timetable in that modification
- Set the starting time of that modification to something other than 7AM (e.g. 9AM)
- Create another modification of another type, e.g. "Adjust dwell time"
- Switch back to your "add trips" modification
- The starting time of your timetable will be reset to 7AM even though you formerly set it to 9AM

trevorgerhardt:
fixes #231

### DIFF
--- a/lib/components/modification/timetable-entry.js
+++ b/lib/components/modification/timetable-entry.js
@@ -29,9 +29,8 @@ export default class TimetableEntry extends Component {
   }
 
   /** set the start time of this modification */
-  setStartTime = (e) => {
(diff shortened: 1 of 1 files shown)


mattwigway:
Oof, how did this one get by?

PR #252, opened 2017-01-26
Filter modifications

mattwigway:
Allow filtering modifications using search text.

### DIFF
--- a/lib/components/modification/list.js
+++ b/lib/components/modification/list.js
@@ -5,7 +5,7 @@ import Icon from '../icon'
 import * as types from '../../utils/modification-types'
 import ModificationGroup from './group'
 import VariantEditor from '../../containers/variant-editor'
-import {Group, Text} from '../input'
+import {Input} from '../input'
 import messages from '../../utils/messages'
 
 export default class ModificationsList extends DeepEqualComponent {
@@ -95,12 +95,15 @@ export default class ModificationsList extends DeepEqualComponent {
           </a>
         </div>
 
-        <Group label={messages.modification.filter}>
-          <Text
+        <div className='block'>
+          <Input
+            aria-label={messages.modification.filter}
+            placeholder={messages.modification.filter}
             onChange={this._setFilter}
+            type='text'
             value={filter || ''}
-            aria-label={messages.modification.filter} />
-        </Group>
+            />
+        </div>
 
         {modificationsByType && Object.values(types).map((type) => {
           if (typeof type !== 'string') return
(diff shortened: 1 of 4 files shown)


mattwigway:
I can't figure out quite how to get the padding right:

image

lib/components/modification/list.js
```diff
@@ -84,14 +95,22 @@ export default class ModificationsList extends DeepEqualComponent {
           </a>
         </div>
 
+        <Group label={messages.modification.filter}>
+          <Text onChange={this._setFilter} value={filter || ''} />
+        </Group>
+
         {modificationsByType && Object.values(types).map((type) => {
           if (typeof type !== 'string') return
+          const modifications = modificationsByType[type]
```
trevorgerhardt: What do you think about filtering the modifications when we're setting the new state? Then use `filteredModificationsByType` when rendering.

mattwigway:
Please! I'm thinking we should also add a clear filter button.

On Jan 27, 2017 8:40 AM, "Trevor Gerhardt" <notifications@github.com> wrote:

*@trevorgerhardt* commented on this pull request.

Mind if I take a stab at the padding in this PR?
------------------------------

In lib/components/modification/list.js
<https://github.com/conveyal/analysis-ui/pull/252#pullrequestreview-18839376>
:

>          {modificationsByType && Object.values(types).map((type) => {
           if (typeof type !== 'string') return
+          const modifications = modificationsByType[type]

What do you think about filtering the modifications when we're setting the
new state? Then use filteredModificationsByType when rendering.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/conveyal/analysis-ui/pull/252#pullrequestreview-18839376>,
or mute the thread

lib/components/modification/list.js
mattwigway: That seems like a good plan.

mattwigway: This will also make it easier if in the future we want to make the search smarter (e.g. it could be useful to also search the names of affected transit routes, we don't have to put all of that logic in the `render()` function.

lib/components/modification/list.js
```diff
@@ -21,7 +23,8 @@ export default class ModificationsList extends DeepEqualComponent {
   }
 
   state = {
-    showAllOnMap: true
+    showAllOnMap: true,
+    filter: ''
```
trevorgerhardt: Btw @mattwigway, whenever possible please order alphabetically. If grouping or other structure is warranted, group and order within the group. 

Long lists of unordered variables or imports can become quite difficult to read through.

mattwigway: agreed. I assume you're aware of https://github.com/atom/sort-lines ?

PR #258, opened 2017-01-27
Pull out common code between Add Trips & Reroute.

trevorgerhardt:
The form and functionality required to begin editing the alignments of Add Trips & Reroute modifications is almost identical and therefore an appropriate place to pull out into a common component.

### DIFF
--- a/configurations/default/messages.yml
+++ b/configurations/default/messages.yml
@@ -35,6 +35,9 @@ scenario:
   editTitle: Edit scenario
   route: Route
 transitEditor:
+  startEdit: Edit route geometry
+  stopEdit: Stop editing
+  bidirectional: Bidirectional
   followRoad: Follow streets
   makeStop: make stop
   makeControlPoint: make control point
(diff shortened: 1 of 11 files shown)


trevorgerhardt:
@mattwigway looks like I messed up the defaults. I believe it is working correctly now, please take a look to verify the behavior!

mattwigway:
Not quite. When I edit an add-trip-pattern modification, the checkbox now reflects the behavior correctly, but defaults to extending from the beginning. We should default to extending from the end. Also, when I edit a reroute that has only a from stop, it extends from the beginning rather than the end.

trevorgerhardt:
@mattwigway so the logic all works the same on my end. This latest change makes it more obvious where we're getting the `extendFromEnd` boolean from since it's different for reroute and add trip patterns.

Please test again!

mattwigway:
Great! I'll take a look before I take off today, unless you were looking for a quick test right away.

mattwigway:
It's working locally now.  Weird 🌔 .

abyrd:
Actually the moon phase is more like 🌒 .

PR #270, opened 2017-02-01
Lng to Lon

trevorgerhardt:
WARNING: Renaming a lot of variables so high potential for breaking things.

Update to use [@conveyal/lonlat](/conveyal/lonlat) and rename all variables/function names/parameters from `lng` to `lon`. Only instances we'll still see `lng` is when using Leaflet directly. Note: `lonlat` does produce leaflet compatible output (see [Leaflet/LatLng.js](https://github.com/Leaflet/Leaflet/blob/master/src/geo/LatLng.js#L127)).


lib/components/application.js
```diff
@@ -232,8 +232,8 @@ function getRouteBounds (route) {
   const geometries = route.patterns.map((pattern) => pattern.geometry)
   const coordinates = geometries.reduce((coordinates, geometry) => coordinates.concat(geometry.coordinates), [])
   if (coordinates.length > 0) {
-    const bounds = latLngBounds(lonlng(coordinates[0]))
-    coordinates.forEach((coordinate) => bounds.extend(lonlng(coordinate)))
+    const bounds = latLngBounds(lonlat(coordinates[0]))
```
mattwigway: Is latLngBounds the right function name here?

trevorgerhardt: Yup, `latLngBounds` comes from Leaflet.

Issue #271, opened 2017-02-01
Speeds show up as NaN in report


mattwigway:
I think this is due to us changing to use segmentSpeeds. It looks like the snapshots caught this but were updated to include the NaN.

PR #272, opened 2017-02-01
Fix report speeds

mattwigway:
Correct segmentSpeeds showing up as NaN (#271).

### DIFF
--- a/lib/components/modification/segment-speeds.js
+++ b/lib/components/modification/segment-speeds.js
@@ -126,7 +126,7 @@ export default class SegmentSpeeds extends DeepEqual {
 
     const averageSpeed = getAverageSpeedOfSegments({
       segmentDistances,
-      speeds: segmentSpeeds
+      segmentSpeeds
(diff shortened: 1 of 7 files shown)


lib/components/modification/segment-speeds.js
```diff
@@ -185,9 +185,9 @@ export default class SegmentSpeeds extends DeepEqual {
   }
 }
 
-function getAverageSpeedOfSegments ({segmentDistances, speeds}) {
+export function getAverageSpeedOfSegments ({segmentDistances, segmentSpeeds}) {
```
trevorgerhardt: I'd prefer exporting only Components from the components directory and moving functions like this into utils. 

Utils could probably use a bit of organization soon.

trevorgerhardt: Other than that 👍

mattwigway:
@trevorgerhardt should be good to go!

PR #277, opened 2017-02-02
Utilize woonerf bootstrapping

trevorgerhardt:


lib/routes.js
```diff
@@ -0,0 +1,58 @@
+/**
+ * Main entry point for the scenario editor
```
mattwigway: Comment needs updating.

lib/utils/auth0.js
```diff
@@ -7,7 +7,7 @@ import messages from './messages'
```
mattwigway: Doesn't/shouldn't woonerf handle a lot of this auth0 boilerplate?

lib/utils/authenticated-fetch.js
```diff
@@ -2,13 +2,19 @@ import dbg from 'debug'
 import fetch from 'isomorphic-fetch'
```
mattwigway: I thought we were eliminating authenticatedFetch?

lib/utils/auth0.js
trevorgerhardt: Wanted the changes to be incremental so we're keeping this Auth0 util for now. Will change that in the future.

lib/utils/authenticated-fetch.js
trevorgerhardt: Same as the last comment. Will remove this in the future.

lib/routes.js
trevorgerhardt: Updated 👌

PR #278, opened 2017-02-06
Remove internal authenticated fetch

trevorgerhardt:
Anytime authenticated fetch is used manually we are unable to map it easily to state to show loading indicators or status messages. This PR is going to move all uses of fetch to use the woonerf/fetch action pattern. 

Note: This does not include edits to `utils/browsochrones.js`. That requires a lot of unpacking and will come in a subsequent pull request.

### DIFF
--- a/__mocks__/isomorphic-fetch.js
+++ b/__mocks__/isomorphic-fetch.js
@@ -1,0 +1,5 @@
+const ifetch = require('isomorphic-fetch')
+
+module.exports = (url, opts) => url[0] === '/'
+  ? ifetch(`http://mockhost.com${url}`, opts)
+  : ifetch(url, opts)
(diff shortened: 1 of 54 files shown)


mattwigway:
Bundle upload does not work with these changes:

image

Grid upload is broken as well, but there are no informative error messages.

Also, please rebase; some issues that are fixed in dev regarding the switch to d3-color are still present here.

trevorgerhardt:
@mattwigway fixes have been made, please check again when you have a minute!

mattwigway:
Bundle upload appears to still not be working. I no longer see an error, but it remains spinning "processing" even once the polled bundle status is DONE.

mattwigway:
Grid upload works though!

trevorgerhardt:
@mattwigway regarding changes to `utils/segments`... my comments yesterday about merging/rebasing were referring to this PR in that it caused an unusual replay of commits to occur. I think everything is fine, it just appears like I changed more in this PR than I actually have.

trevorgerhardt:
Good catch on the bundle upload, it was a simple fix: had checked "state" instead of "status". That's what these reviews are for! 👌

mattwigway:
Works now, thanks @trevorgerhardt!

PR #291, opened 2017-02-14
Fix reroute nesting

### LINKED ISSUE #288: Cannot deep link to a reroute modification
If you reload the page with a reroute modification open, the modification does not load:

image

This error appears in the console:

image

trevorgerhardt:
closes #288


trevorgerhardt:
@mattwigway I added a mocked feed to the default mocked store that our tests use.

PR #295, opened 2017-02-15
Prefer const where possible

trevorgerhardt:
All this took was `mastarm lint --fix` using [this](https://github.com/conveyal/mastarm/tree/add-prefer-const) branch of mastarm. ESLint ftw.


trevorgerhardt:
🙏

Issue #298, opened 2017-02-17
Add field for how much time to remove from schedule around removed stops


mattwigway:
See also conveyal/r5#237

PR #307, opened 2017-02-23
Remove time when removing stops

mattwigway:
Allow removing time when removing stops. Also add ability to display scenario application warnings.


mattwigway:
Fixes #298

Issue #300, opened 2017-02-22
UI lock screen is blank


mattwigway:
When the UI is locked with an error, no text appears. I noticed this a while back but hadn't had time to ticket it.

![image](https://cloud.githubusercontent.com/assets/566958/23222402/ffea217e-f8e4-11e6-9797-16cf99ac5240.png)

mattwigway:
I'm seeing this in analysis mode.

PR #308, opened 2017-02-24
Pass the error value to lockUiWithError

trevorgerhardt:
Which is parsed via woonerf

addresses #300

@mattwigway is there a specific example you had that I can test to make sure it works for your case?

### DIFF
--- a/lib/actions/get-feeds-routes-and-stops.js
+++ b/lib/actions/get-feeds-routes-and-stops.js
@@ -32,7 +32,7 @@ function getAllRoutesAndStops ({
     url: query.compose(query.data, {bundleId, routeIds}),
     next (error, response) {
       if (error) {
-        return lockUiWithError(error)
+        return lockUiWithError(error.value)
       }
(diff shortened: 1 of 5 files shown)


mattwigway:
@trevorgerhardt I'm still seeing the same blank error box.

The way I make it happen reliably is to add the line `halt(400);` on line 33 of SinglePointAnalysisController in the analyst project, which makes the backend always crash.

trevorgerhardt:
@mattwigway this should fix other error messages so we should move the "Display error by manually adding halt" to another task.

mattwigway:
@trevorgerhardt I don't fully understand. #300 was fixed by https://github.com/conveyal/analysis-ui/pull/312 ; what value add does this provide?

trevorgerhardt:
Ah! 👍

mattwigway:
hmm, some errors are showing the blank error box again, specifically if the backend crashes when editing a project. I'm wondering if we need this change after all.

PR #312, opened 2017-02-24
fix(error): Fix lock UI with error.

### LINKED ISSUE #300: UI lock screen is blank

mattwigway:
Allow calling lock ui with error with a string, an Error object, or { error, detailMessage } object.
Don't use createAction due to an upstream :ant: https://github.com/acdlite/redux-actions/issues/194

fixes #300

### DIFF
--- a/lib/actions/analysis/index.js
+++ b/lib/actions/analysis/index.js
@@ -1,8 +1,8 @@
 import lonlat from '@conveyal/lonlat'
 import {createAction} from 'redux-actions'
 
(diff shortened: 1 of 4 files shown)


mattwigway:
@trevorgerhardt I'm going to go ahead and merge so I can cut 1.1, but interested to hear your comments.

PR #326, opened 2017-03-07
feat(regional): Allow using an uploaded grid to constrain the extents of a regional analysis

### LINKED ISSUE #325: Set different origins for regional analysis
Currently we use all origins within the project bounds for regional analyses. Sometimes the project bounds are extremely large, and we only want to perform the analysis for a small part of the project; we should allow restricting the extents.

mattwigway:
This is one way to fix #325; it allows the user to select a grid (presumably one they've uploaded) and it will use its extents as the extents of the regional analysis. This adds a flask button that causes a grid selector to be shown:

image

I'm not sure if it's the best way since it requires uploading a grid with the proper extents, but I needed to hack something together for a project and this was the easiest. Now that we have a bit more time I'm open to suggestions for better implementations.

Just drawing a box on the map is attractive but we need to save the extents somehow so multiple analyses can use the same extents.

### DIFF
--- a/lib/components/buttons.js
+++ b/lib/components/buttons.js
@@ -1,6 +1,7 @@
-import React from 'react'
+import React, {PropTypes} from 'react'
 
-import {pure} from './deep-equal'
+import DeepEqualComponent, {pure} from './deep-equal'
+import Icon from './icon'
 
 export const Group = pure(function ButtonGroup ({children, justified, vertical, radio}) {
   const classNames = ['btn-group']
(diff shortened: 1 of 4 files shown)


lib/components/buttons.js
```diff
@@ -54,3 +55,46 @@ export const Button = pure(function Button ({block, children, className, onClick
     {...props}
     >{children}</a>
 })
+
+/** A toggle button that uses a shaded icon to indicate checkiness */
+export class IconToggle extends DeepEqualComponent {
```
trevorgerhardt: Let's put this in it's own file instead of in buttons.js

mattwigway:
I will pull the IconToggle class out of this PR, but we are going to implement constraining regional analysis origins a different way.

PR #334, opened 2017-03-16
build(deps): bump gridualizer to 2.0.1 to avoid import/constructor er…

### LINKED ISSUE #328: _xorshift2.default is not a constructor
I think we need to bump the gridualizer dependency to fix this.

mattwigway:
…ror :tools:

the gridualizer quantile scale did not work, due to importing the default object instead of the
constructor from XorShift.

Fixes #328.

I suppose to do this right we should really update package.json as well, but I was under the mistaken impression that if I pushed a new patch release I wouldn't have to update anything. The content of this branch is what is deployed now (actually the contents of this branch cherry-picked onto master).

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3,8 +3,8 @@
 
 
 "@conveyal/gridualizer@^2.0.0":
-  version "2.0.0"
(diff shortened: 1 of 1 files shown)


mattwigway:
Actually, rebasing with yarn.lock is a bad idea.

PR #343, opened 2017-03-20
Update dependencies & snapshots

trevorgerhardt:


lib/components/__tests__/__snapshots__/input.js.snap
```diff
@@ -1,10 +1,16 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
 exports[`Component > Input Checkbox renders correctly 1`] = `
 <div
-  className="checkbox">
-  <label>
+  className="checkbox"
+>
+  <label
+    htmlFor="Do the thing?"
```
mattwigway: This seems like not how it's supposed to work?

lib/components/__tests__/__snapshots__/input.js.snap
```diff
@@ -13,8 +19,11 @@ exports[`Component > Input Checkbox renders correctly 1`] = `
 
 exports[`Component > Input File renders correctly 1`] = `
 <div
-  className="form-group">
-  <label>
+  className="form-group"
+>
+  <label
+    htmlFor="Select files"
```
mattwigway: same comment.

lib/components/analysis/spectrogram.js
```diff
@@ -82,8 +82,8 @@ export default class Spectrogram extends Component {
       maxAccessibility
     } = this.props
 
-    width = width - width % MINUTES // make nice integer divisors
-    height = height - height % bins
+    width = width - (width % MINUTES) // make nice integer divisors
```
mattwigway: Is assuming default order of operations now forbidden?

lib/components/date-picker.js
```diff
@@ -22,7 +22,7 @@ export default class DatePicker extends DeepEqualComponent {
 
   render () {
     const {value, label, ...rest} = this.props
-    return <label>
+    return <label htmlFor={label}>
       {label}
```
mattwigway: I don't think this is the correct use of htmlFor. I think it should contain the ID of the element that it is referring to.

lib/components/scenario-map/stop-layer.js
```diff
@@ -45,10 +45,8 @@ export default class StopLayer extends Component {
       <FeatureGroup>
         {routeStops
           .map((stop) => {
-            const stopIsSelected = nullIsWildcard && // TODO: not this
-              modification.stops == null ||
-              modification.stops != null &&
-              modification.stops.indexOf(stop.stop_id) > -1
+            const stopIsSelected = (nullIsWildcard && modification.stops == null) ||
```
mattwigway: Looks like we lost a comment here, intentional?

lib/components/analysis/spectrogram.js
trevorgerhardt: We base our linting off the standardjs ruleset and they added the `no-mixed-operators` rule recently: https://github.com/feross/standard/blob/master/CHANGELOG.md#new-rules-1

lib/components/scenario-map/stop-layer.js
trevorgerhardt: Yes, I couldn't remember exactly why I put it there other than that I don't like that kind of variable.

trevorgerhardt:
@mattwigway was going for a quick and dirty way to pass the a11y linter rules. Since we're using react components it shouldn't be too hard to automate this in a smarter way. I'll look into a solution that uses an id if passed or adds an id to the input based off the name/label and use that as htmlFor also.

lib/components/buttons.js
```diff
@@ -35,7 +35,10 @@ export const Button = pure(function Button ({block, children, className, onClick
   if (radio && checked) classNames.push('active')
 
   if (radio) {
-    return <label htmlFor='Button' className={classNames.join(' ')}>
+    return <label
+      htmlFor={props.id || props.name || 'radio-button'}
+      className={classNames.join(' ')}
+      >
       <input
         type='radio'
```
mattwigway: Doesn't the input need to have an `id` to match the label? Also, `radio-button` is not a good id as it probably won't be unique.

lib/components/input.js
```diff
@@ -186,3 +182,11 @@ export class SelectMultiple extends StateToValue {
     )
   }
 }
+
+function propsToId ({
+  id,
+  label,
+  name
+}) {
+  return id || toKebabCase(`${name || label}-input`)
```
mattwigway: Won't we wind up with a lot of duplicated IDs in repeated elements (e.g. frequency entries).

lib/components/buttons.js
trevorgerhardt: I think limited, duplicated ids until we properly name everything is ok. I'd rather not use uuids or [this approach](https://github.com/conveyal/analysis-ui/pull/343/commits/ba81f89143924bff6072b9767d1a998f11a69c68#diff-f2636176b96b5e2bf2ed574769b38fb0) across our components.

mattwigway: Yeah, I get the desire to avoid id-generation hacks. But I think duplicate IDs are going to be worse than no IDs at all for screen readers, which kind of defeats the purpose of using the a11y linter at all.

trevorgerhardt:
@mattwigway these latest changes should produce unique ids every time. Ideally we could do it differently, but I think this works well based off what the snapshots produce.

trevorgerhardt:
@mattwigway I believe that is the case.

PR #347, opened 2017-03-30
Regional analysis bounds

mattwigway:
This allows creating regional analyses with bounds that are not the same as the project bounds.


mattwigway:
When creating a regional analysis, there are bounds displayed on the map. You can edit them, or select an existing regional analysis to match the bounds.

Issue #318, opened 2017-02-25
Phasing UI


mattwigway:
We had an issue for this in Asana but it never made it over here.

We can now in r5 specify phasing of frequency entries relative to each other (e.g. you have a pulse, or you have multiple branching lines that are scheduled to be exactly out of phase on a common trunk). We need to break this out into the UI. There are two cases:

1) Phase of entry B is relative to a newly created frequency entry A. In this case, entry A should have an entryId field, and entry B will have a phaseFromTimetable field referring to A.
2) Phase of entry B is relative to an existing frequency entry A. In this case, entry B will have a phaseFromTimetable field referring to the ID of A, which is "<trip id>_hh:mm:ss_to_hh:mm:ss_every_<mm>m<ss>s", e.g. "WMATA:GLENMONT_07:00:00_to_11:00:00_every_02m00s"

In either case, entry B will have a phaseFromStop attribute, which is the stop in the pattern of entry A from which the phase is determined. It will also have a phaseAtStop attribute, which is the stop in entry B where the phasing occurs. Finally, there is a phaseSeconds attribute, which is a nonnegative number of seconds from when a vehicle on entry A leaves the phaseFromStop to when a vehicle on entry B leaves the phaseAtStop.

So for example: suppose that blue line is going to leave Farragut West four minutes after the red line leaves Farragut North. The blue line would have the following in its frequency entry:

phaseFromStop: "WMATA:FarragutNorth",
phaseAtStop: "WMATA:FarragutWest",
phaseFromTimetable: "WMATA:GLENMONT_07:00:00_to_11:00:00_every_02m00s",
phaseSeconds: 600

If the red line was specified in the scenario rather than from GTFS, it might look like this, assuming the relevant frequency entry on the red line had an entryId of "redLineEast":

phaseFromStop: "WMATA:FarragutNorth",
phaseAtStop: "WMATA:FarragutWest",
phaseFromTimetable: "redLineEast",
phaseSeconds: 600

Currently trips can only be phased at existing stops. We'll relax that eventually.

From the UI side, I'd recommend just generating an ID for every frequency entry, probably using a UUID. For now it would be fine to only support selecting a newly-created frequency entry as the phaseFromTimetable; using existing frequency entries requires getting a lot of data into the UI.

This will require some changes in the analyst project in the models. It should be as simple as just adding the string fields to the appropriate models.

[...]

PR #354, opened 2017-04-12
Phasing

trevorgerhardt:
Addresses #318 for Convert to Frequency modifications. Shouldn't be too hard to phase Add Trip Modifications to a Frequency Entry, just need to figure out how to handle stop selection in that case.


mattwigway:
@trevorgerhardt can't we just use the graphical stop selector? I'm thinking we want to do this at all times, frequently the stop names are empty or not meaningful.

mattwigway:
i.e. the same one we use for phasing?

mattwigway:
i.e. the same one we use for reroutes?

lib/actions/__tests__/index.js
```diff
@@ -63,15 +63,11 @@ describe('actions > index', () => {
     })
   })
 
-  describe('> map state', () => {
```
trevorgerhardt: Meant to just group the user action.

lib/utils/timetable.js
```diff
@@ -1,6 +1,9 @@
+import format from 'date-fns/format'
```
trevorgerhardt: Ideally I'd like us to move away from `moment.js` and use `date-fns` as much as possible.

lib/components/modification/phase.js
```diff
@@ -0,0 +1,94 @@
...
+          <Select
+            name='Phase from stop'
+            onChange={this._setPhaseFromStop}
+            options={selectedPhaseFromTimetableStops.map((stop) =>
+              ({value: stop.stop_id, label: stop.stop_name}))}
+            placeholder='Select stop to phase from'
+            value={phaseFromStop}
+            />
+        </FormGroup>}
+      <InputNumber
```
trevorgerhardt: 👌

trevorgerhardt:
@mattwigway 

1. Stop selection with the GUI wasn't as straightforward of an implementation due to the nature of the selection but it can be done. The feed I was using had stop names so it appeared to be an appropriate first pass.
2. Didn't even think that you could phase to another FE within the modification 👌 
3. If we can't phase to new stops then we can't phase Add Trips modifications since they're all new stops...right? Or is it ok to phase AT a new stop FROM an existing stop? Cause then we could do it. 
4. 👌

mattwigway:
1. Okay, yeah, let's go ahead and leave the stop names for now
3. Sorry, I was unclear. It's totally possible to phase add trips modifications (even to entries in the same modification), but you have to do it at stops which are present in the GTFS (i.e. the blue ones on the map, that have a stop ID and by extension a stop name, so the stop names approach works for now). The blue stops on the map indicate that the added trip stops at an existing stop, the black stops indicate a brand new stop (which we can't use for phasing yet) and the gray stops autogenerated stops based on the auto-create stops distance (which we should never phase to, since they move around)

lib/utils/timetable.js
mattwigway: :+1:

trevorgerhardt:
@mattwigway latest changes should now work with Add Trips modifications (with a corresponding update to the phasing branch in Analysis Server)

mattwigway:
We should have a warning when you phase from an entry that doesn't have the same headway as the entry you're phasing. It's not necessarily wrong (many agencies have all lines arrive at a central location at the top of an hour, and then more frequent lines also arrive during that hour), but it is often a mistake. For instance, if you phase one line that runs every 12 minutes and another that runs every 15, they will only line up once per hour—probably not what was intended.

mattwigway:
Nice work!

The "phase from stop" seems to be missing on add trips modifications. I'm also thinking the phasing stuff takes up a lot of screen real estate, maybe we should have a checkbox that turns it all on and off?

We also need to modify the R5 exporters.

trevorgerhardt:
`Phase from timetable` will only show up when a `Phase at stop` is selected. Similarly with `Phase from stop` to `Phase from timetable`. We could apply that to `Phase minutes` also so that only one field is showing until it is in use.

^ likewise I imagine `Phase from stop` is not appearing because there are no stops available to phase from. This could be a code error, I'll add a warning and investigate.

mattwigway:
Yeah, let's hide phaseMinutes.

mattwigway:
Still not seeing the phase from stop tho:

image

Also, it occurs to me that any entry that has "exact times" selected should not be available for phasing (either as a from or to entry).

mattwigway:
Looking good, although it looks like it's not possible to select the first stop as the phase from stop?

mattwigway:
A few more comments on the R5 exporter:

- The `phaseAtStop` and `phaseFromStop` need to be feed-ID-qualified (e.g. they need to say `MBTA:5410` rather than `5410`)
- the `id` field on a timetable needs to be exported as `entryId`

mattwigway:
I fixed the `entryId` issue and removed the use of the spread operator, but I'm not sure what to do about the feed-qualified stop IDs. We can't do the same thing we do with the source trips, using the feed of the modification, because the stops we are phasing with may come from a different feed.

mattwigway:
This is _so close_, thanks for all your work on this @trevorgerhardt! I'm seeing the `phaseFromStops` in adjust frequency modifications being qualified twice, e.g. I'm getting `MBTA:MBTA:64` as `phaseFromStop`.

mattwigway:
With the latest changes, it looks like `phaseSeconds` is not being exported. Additionally it looks like the phase from timetables are being named incorrectly:

image

The timetable in question runs from 7:00 AM until 10:00 PM, not sure why it's saying 2am to 5pm.

lib/selectors/modifications.js
```diff
@@ -0,0 +1,4 @@
+
+export default function selectModifications (state) {
+  return state.scenario.modifications
```
mattwigway: What's the advantage to using a selector here?

lib/components/modification/adjust-dwell-time.js
```diff
@@ -35,34 +38,43 @@ export default class AdjustDwellTime extends Component {
...
+            selectedStops={selectedStops}
             update={update}
             />}
 
-        <br />
-
         <div className='form-group'>
-          <label htmlFor='adjust-dwell-time-scale' className='radio-inline'>
-            <input id='adjust-dwell-time-scale' type='radio' value='scale' checked={modification.scale} onChange={this._setScale} />Scale existing dwell times by
+          <label htmlFor='adjust-dwell-time-scale'>
```
trevorgerhardt: Yup!

lib/components/modification/editor.js
```diff
@@ -152,51 +116,70 @@ export default class ModificationEditor extends Component {
...
+  setVariant
+}) {
+  return <div>
+    <legend>Active in variants</legend>
+    <div className='form-inline'>
+      {allVariants.map((v, i) => <Checkbox
+        checked={activeVariants[i]}
+        label={i + 1}
+        key={`variant-${i}-modification-${modificationId}`}
+        onChange={(e) => setVariant(i, e.target.checked)}
```
trevorgerhardt: Totally. An easy/clearer way of doing it wasn't immediately apparent

lib/selectors/modifications.js
trevorgerhardt: Standardization of how we "retrieve" it. Could change in the future, but the input and output will be the same.

trevorgerhardt:
🎉🎊🎉

PR #356, opened 2017-04-13
fix(buttons): initial pass at making radio buttons work again

mattwigway:
This fix works, however the main issue is that to toggle a button you now have to click on the text, not the button itself. @trevorgerhardt any thoughts?

### DIFF
--- a/lib/components/__tests__/__snapshots__/buttons.js.snap
+++ b/lib/components/__tests__/__snapshots__/buttons.js.snap
@@ -1,151 +1,76 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Component > Buttons Button radio buttons work 1`] = `
-<ButtonGroup
-  radio={true}
->
+<Group>
   <div
     className="btn-group"
-    data-toggle="radio"
   >
-    <Button
-      name="radio"
+    <RadioButton
+      id="r0"
+      name="radio-group"
       onChange={[Function]}
-      radio={true}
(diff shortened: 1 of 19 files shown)


trevorgerhardt:
Is there an instance where we use `children` inside of a radio?

trevorgerhardt:
The controlled element (`input`) is allowed inside of a `label`. Nothing else is.

trevorgerhardt:
Also, is it possible to create a test case for this that breaks with the current code in dev? That would be preferable for this kind of a bug fix. That way if we update the way we implement radios in the future we'll be able to detect breakages.

mattwigway:
I'm not quite sure how to implement a test, because I think the way it is
currently implemented in dev should work. However, clicking on the labels
containing inputs does not toggle them in chrome.

trevorgerhardt:
We're not currently snapshot testing the radio style of buttons at all yet, so that would be a good start. Then I would simulate clicks on various parts of the HTML (label, input) and make sure the snapshots change correctly according to the clicks. See [here](https://github.com/conveyal/analysis-ui/blob/dev/lib/components/__tests__/collapsible.js#L24) for an example of simulating clicks.

This will at least get to the heart of if it's our implementation error vs differences in browsers.

mattwigway:
That last commit adds a failing test, I'm not sure quite where to go from here.

trevorgerhardt:
👌 👌 great start. I can try to take a look on Monday.

PR #370, opened 2017-05-09
Regional analysis percentiles

mattwigway:
Allow setting the percentile of a regional analysis. Depends on conveyal/analysis-backend#27.


configurations/default/messages.yml
```diff
@@ -104,6 +105,7 @@ analysis:
   errorsInScenario: Errors were found in the scenario:
   warningsInScenario: Warnings were found applying the scenario
   errorsInModification: in modification %s
+  travelTimePercentile: Travel time percentile (reliab)
```
mattwigway: Nope, not abbreviated on purpose, oops.

mattwigway: Fixed, I'm not sure why we'd want to move this to the top level though.

trevorgerhardt: No reason now! Just need to update the tests then.

mattwigway:
@trevorgerhardt I think this is done, turns out the failing tests were showing an issue in the code.

Issue #320, opened 2017-02-27
Cannot enter decimal point in number fields


mattwigway:
For example the frequency field in a timetable. I think this is because onChange is getting called with e.g. `4.` which is converting back to `4`. We should probably not call onChange when the number entered ends with a `.`

abyrd:
@marcovanderlinden just run into this again. It's making it harder for him to do some work. @trevorgerhardt any chance you can prioritize this one?

PR #381, opened 2017-05-17
Add mm:ss component

trevorgerhardt:
And switch Phasing minutes, Dwell time, and segment speeds to use it.

refs #320

### DIFF
--- a/lib/components/minutes-seconds.js
+++ b/lib/components/minutes-seconds.js
@@ -32,12 +32,22 @@ export default class MinutesSeconds extends PureComponent<void, Props, void> {
 
   _onChange = (e: Event) => {}
 
+  _onKeyDown = (e: Event & {
+    currentTarget: HTMLInputElement,
+    which: number
+  }) => {
+    if (e.which === 13) {
+      this.props.onChange(stringToSeconds(e.currentTarget.value))
+    }
+  }
+
   render () {
     const {seconds, ...props} = this.props
     return <Text
       {...props}
       onBlur={this._onBlur}
       onChange={this._onChange}
+      onKeyDown={this._onKeyDown}
       units='mm:ss'
       value={seconds > 0 ? secondsToMmSsString(seconds) : ''}
       />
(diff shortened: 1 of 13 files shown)


trevorgerhardt:
Allows time to be entered in with a decimal point also.

trevorgerhardt:
@MarcovanderLinden @abyrd which inputs were you having trouble with? I can switch them to use this new component

MarcovanderLinden:
At the moment I specifically have trouble entering frequency entries with decimal points (e.g. 7,5 minutes) this is in all modification types that ask for a frequency.

abyrd:
@marcovanderlinden that's probably an internationalization issue. Please use the English decimal separator (a dot/full stop).

MarcovanderLinden:
@abyrd , that's not it, i've tried 7,5 7.5 and 7:5, all of these show 75 or 0. I think that what @mattwigway said on slack makes sense: the system checks input really fast and doesn't recognize 7. and before i can type the 5 it turns the value to 0. 
His workaround works fine by the way (copying the number 7.5 in)

trevorgerhardt:
@MarcovanderLinden this pull request should address that issue. We still need to switch the frequency input to use the new component and then deploy the fix to production.

lib/components/minutes-seconds.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  render () {
+    const {seconds, ...props} = this.props
+    return <Text
+      {...props}
+      onBlur={this._onBlur}
+      onChange={this._onChange}
+      onKeyDown={this._onKeyDown}
+      units='mm:ss'
```
mattwigway: This needs to be a message so we can internationalize it.

lib/components/modification/segment-speeds.js
```diff
@@ -172,12 +167,10 @@ export default class SegmentSpeeds extends DeepEqual {
                 value={speed}
                 units='km/h'
                 />
-              <Input
+              <MinutesSeconds
                 placeholder='Time'
```
mattwigway: This should also be internationalized.

PR #383, opened 2017-05-19
Deep-six browsochrones

mattwigway:
Remove browsochrones and use much simpler travel time grids.


mattwigway:
@trevorgerhardt this is about ready, I fixed all the existing tests but still want to add a few new ones.

mattwigway:
This should be good to go. I'm not sure why the codecov went down so much, I don't think I did that.

mattwigway:
Wait wait, I broke a test.

mattwigway:
Fixed the tests, this should be good to go @trevorgerhardt!

lib/actions/analysis/index.js
```diff
@@ -41,149 +56,269 @@ export const setComparisonScenario = ({ id, bundleId, variantIndex }) => [
...
-        } else {
-          // unexpected error, rethrow
-          throw err
+export const setCurrentIndicator = (projectId, indicator) => fetch({
+  // Chrome 57 preserves headers when following redirects, in a departure from previous versions
+  // this means the auth header from authenticatedFetch is passed to S3, which causes S3 to fail.
+  // Instead we request the URL from the server as JSON and then follow manually
+  url: `/api/grid/${projectId}/${indicator}?redirect=false`,
+  next: (err, res) => {
+    if (err) {
```
trevorgerhardt: As you have now seen in my latest PR manually doing `lockUiWithError` is not necessary when using `fetch` and this exact format will result in a blank locked modal.

lib/components/analysis/stacked-percentile.js
```diff
@@ -0,0 +1,409 @@
...
+          {text}
+        </text>)}
+    </g>
+  }
+
+  renderXAxis (): Element {
+    const {height} = this.props
+    return <g transform={`translate(0 ${height})`}>
+      {/* We don't explicitly center the '15 minutes' text; fudge the left side a bit so that the 15 appears centered */}
+      {[15, 30, 45, 60, 75, 90, 105, 120].map(v => <text
```
trevorgerhardt: I was going to save this to the end, but arrays like this and random numbers sprinkled throughout code (like the 15 and 120 below this) should be hoisted as named constants at the top of the file and if it's something that's possibly common to more than one file moved to the `constants/` directory.

lib/selectors/accessibility.js
```diff
@@ -0,0 +1,40 @@
...
+      if (travelTimeSurface.get(travelTimeX, travelTimeY)[POSITION] < isochroneCutoff) {
+        accessibility += destinationGrid.data[y * destinationGrid.width + x]
+      }
+    }
+  }
+
+  return accessibility
+}
+
+export const accessibility = createSelector(
```
trevorgerhardt: Let's stick to a pattern of exporting one selector per file. Along with being a common pattern amongst all the other selectors so far it allows us to do:

```js
import selectAccessibility from '../selectors/accessibility'
```

The `computeAccessibility` function can be moved to `utils` and tested on it's own there.

lib/components/analysis/stacked-percentile.js
mattwigway: How far do you want to carry this? For example, the box plot code contains a bunch of numbers for the percentage of the width that various components take up - do we want to hoist those as well?

lib/selectors/accessibility.js
mattwigway: Do we really want to move that computeAccessibility function to `utils`? It's not a utility, it's a selector, it's just written this way so I can re-use the same code for the scenario and the comparison.

mattwigway:
@trevorgerhardt I pulled out some magic numbers, and split out the selectors. `utils` seems like a strange place to the functions wrapped by `createSelector` so I left them where they were, but split out the comparison and regular selectors and used default exports. Let me know if that's acceptable.

trevorgerhardt:
Assuming you fix the tests before you merge you're good to go 👌

trevorgerhardt:
@mattwigway looks good now that the tests are passing. Just need to resolve those conflicts 👌

configurations/default/env.yml.tmp
```diff
@@ -5,3 +5,6 @@ LEAFLET_RETINA_URL: https://api.mapbox.com/v4/mapbox.light/{z}/{x}/{y}@2x.png?ac
 LEAFLET_ATTRIBUTION: &copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors, &copy; <a href='https://www.mapbox.com/about/maps/'>Mapbox</a>
 MAPZEN_SEARCH_KEY: test-search-key
 MAPZEN_TURN_BY_TURN_KEY: test-turn-key
+# The minimum version of R5 that the user should be able to select at all
+# v2.2.0 contains the latest API for sending results to the frontend. Previous version will not function.
+MINIMUM_R5_VERSION: v2.2.0
```
trevorgerhardt: This seems like the wrong spot for this. If we created a settings/configuration part of the `store.yml` and a reducer to match it that could take changes later on it would be a better place. 

`env.yml` should be mainly private keys. The `LEAFLET_ATTRIBUTION` is shoved in because that's where the other leaflet data persists but that should probably change too.

lib/components/application.js
```diff
@@ -81,10 +82,11 @@ export default class Application extends Component {
   }
 
   componentWillMount () {
```
trevorgerhardt: Based off [FB React.Component guidelines](https://facebook.github.io/react/docs/react-component.html#componentdidmount) we should do these types of requests in `componentDidMount`.

lib/components/r5-version.js
```diff
@@ -6,79 +6,53 @@ import Select from 'react-select'
...
     onChange: PropTypes.func.isRequired
   }
 
-  state = {
-    allVersions: [],
-    releaseVersions: [],
-    showAllVersions: false
+  componentWillMount () {
+    const {value} = this.props
+    this.setState({ showAllVersions: !RELEASE_VERSION_REGEX.test(value) })
```
trevorgerhardt: This can be done directly while initializing state:

```js
state = {
  showAllVersions: !RELEASE_VERSION_REGEX.test(this.props.value)
}
```

should work.

lib/components/r5-version.js
```diff
@@ -6,79 +6,53 @@ import Select from 'react-select'
...
   }
 
-  componentDidMount () {
-    if (process.env.NODE_ENV !== 'test') this.fetchVersions()
+  toggleAllVersions = () => {
+    this.setState({...this.state, showAllVersions: !this.state.showAllVersions})
```
trevorgerhardt: I know it's still sprinkled throughout the codebase but `this.setState({...this.state, property: value})` is equivalent to `this.setState({property: value})` so passing the rest of the state is unnecessary.

Issue #368, opened 2017-05-03
allow specifying dwell times per-stop in add trips modifications


mattwigway:
This is needed when creating a pulse, and we want all the vehicles to sit at a particular location for five minutes while users transfer between all of them.

Issue #358, opened 2017-04-14
Lock UI with error is still blank in some cases


mattwigway:
For instance, #357

PR #386, opened 2017-05-23
Set dwell times per stop

trevorgerhardt:
* Highlight segments and stops being edited in an Add Trips modification
* Adds ability to set dwell times for each GTFS stop (stops available for phasing)
* Fixes UI Lock error on network requests

Addresses #389, #368, & #358


trevorgerhardt:
This implementation uses an array for dwellTimes of available modification stops. This was chosen because of the initial need (to set dwell times at phase stops).

lib/utils/get-stops.js
```diff
@@ -1,22 +1,26 @@
...
 
 const debug = dbg('utils:get-stops')
 
-export default function getStops (segments) {
+export default function getStops (segments: Segment[]): Stop[] {
   if (segments.length === 0) return []
   if (segments.length === 1 && segments[0].geometry.type === 'Point') {
     const coord = segments[0].geometry.coordinates
     return [{
+      center: coord,
```
mattwigway: Why does every stop have the coords of the first stop as property center?

mattwigway: Also, here this seems to be an array of coordinates from the segment, whereas below it's a single coordinate?

lib/actions/modifications.js
```diff
@@ -84,7 +81,7 @@ const deleteOnServer = (id) =>
     options: {
       method: 'delete'
     },
-    next: (error) => error && lockUiWithError(error)
+    next () {}
```
trevorgerhardt: We don't have a default value for next or check for it's existence so without passing anything we get an error. That should be added.

lib/utils/get-stops.js
trevorgerhardt: When `segments.length === 1` there is only one stop. So "every stop" === "the first stop" in this instance.

trevorgerhardt: And the second comment, good catch 👌

trevorgerhardt: Ah, actually, it's a point so `coordinates` === `Coordinate` here not `Coordinate[]`

lib/components/map/geojson-patterns.js
```diff
@@ -23,7 +26,6 @@ export default class Patterns extends Component {
       data={geometry}
       color={color}
       weight={3}
```
trevorgerhardt: Whoops, yes! I was just experimenting and accidentally left that change in.

mattwigway:
So it's looking pretty good. I really like how the segment you're editing gets highlighted on the map. I just have a few comments:

1) We need to write an R5 exporter for this, the changes don't show up on the R5 side at all.
2) Stops that have been manually created but are not snapped to an existing stop (i.e. do not have a stop ID) do not show up for setting a custom dwell time. I think we should show them and just say "Stop _n_" like we do for the segments—i.e. use the stop name where it's available, but fall back on the number (and the map highlight) when it's not.
3) We need to figure out what to do with automatically created stops. Maybe just leaving them to use the "default" dwell time is fine; the only scenario I'm thinking about is where you have a line that passes through a downtown and a suburb and you want the dwell times to be higher downtown. But maybe if you're going to that level of detail you shouldn't be using automatically created stops at all.
4) I'm wondering if we need to do any special handling when editing geometries for the dwell times. I'm thinking eventually we should move the dwell times and the segment speeds into the segments array, maybe this can wait until then.
5) I am not able to set dwell times at added stops in reroute modifications.
6) I'm also not able to see the added segment on the map.

Let's connect in the AM since we're on a bit of a time crunch with this feature.

trevorgerhardt:
@mattwigway ready for another look. Double check the r5 conversions for me to see if we're doing that appropriately 👌

lib/utils/convert-to-r5-modification/reroute.js
```diff
@@ -53,12 +54,15 @@ export default function convertReroute ({
   const hopTimes = getHopTimes(segmentStops, segmentSpeeds)
 
   // There should be one more dwell time than hop time. The number of hop times depends on fromStop and toStop.
-  const dwellTimes = hopTimes.map((hop) => dwellTime)
-  dwellTimes.push(dwellTime)
+  let realStopIndex = 0
+  const allDwellTimes = segmentStops.map((stop) => {
+    if (stop.autoCreated || !dwellTimes || !dwellTimes[realStopIndex]) return dwellTime
```
mattwigway: (also true in add-trips)

trevorgerhardt: 👌 👌

lib/utils/convert-to-r5-modification/reroute.js
```diff
@@ -56,8 +56,8 @@ export default function convertReroute ({
   // There should be one more dwell time than hop time. The number of hop times depends on fromStop and toStop.
   let realStopIndex = 0
   const allDwellTimes = segmentStops.map((stop) => {
-    if (stop.autoCreated || !dwellTimes || !dwellTimes[realStopIndex]) return dwellTime
-    return dwellTimes[realStopIndex++]
+    if (stop.autoCreated || !dwellTimes) return dwellTime
+    return dwellTimes[realStopIndex++] || 0
```
mattwigway: Doesn't this create an issue if dwellTimes[realStopIndex] is undefined, but the default dwell time is non-zero? For instance, if you have a default dwell time of 30s, won't this cause it to return zero at all stops that don't have a custom dwell time?

mattwigway:
Also, it appears that when the default dwell time is non-zero and some of the specific dwell times are not set, they are exported as zero rather than as the default dwell time. This probably stems from using zero as the flag of 'no dwell time' - if we want to use null we need to change the backend to use an array of `Integer`s (which can be null) as opposed to `int`s (which cannot, and will be coerced to zero if set to null).

trevorgerhardt:
@mattwigway more fun with 0 eh? Switched to a more appropriate comparison. Saving/showing reroute `dwellTimes` should be fixed too. It was saving them, just not passing them correctly to the UI. And`MinutesSeconds` can now handle strings that are empty pre colon.

mattwigway:
It's really close :) when I reload the page, the dwell time at the first stop changes to zero in the UI and the export, regardless of the default travel time, iff I had not set a specific dwell time previously.

PR #410, opened 2017-05-31
Duplicate modification

trevorgerhardt:
screenshot 2017-05-31 18 18 19


trevorgerhardt:
@mattwigway timetable/frequency entry issues still need to be worked out. But every other modification type should copy fine with this.

trevorgerhardt:
@mattwigway ready to be reviewed!

Issue #402, opened 2017-05-29
Copying scenario (using Import Modifications From) button does not copy phase from timetable correctly


mattwigway:
It seems to still work in analysis mode, but the UI does not display a phase from timetable.

trevorgerhardt:
Due to changing modification ids this is a little bit trickier than copying one modification at a time.

mattwigway:
Ah, I didn't realize that the modification IDs were stored in the phase from timetable. Since entry ids are scenario-unique but not globally unique, we could copy all the modifications and preserve their entry IDs, then build a map from entry ID to modification ID, and then update all the phaseFromTimetable information with the new modification IDs.

PR #415, opened 2017-06-02
Fix import modifications

trevorgerhardt:
Correctly match new ids for copied modification timetables when importing from another scenario. Addresses #402

### DIFF
--- a/lib/utils/copy-modifications.js
+++ b/lib/utils/copy-modifications.js
@@ -5,6 +5,12 @@ import {ADD_TRIP_PATTERN, CONVERT_TO_FREQUENCY} from '../constants'
 export default function copyModifications (modifications, newScenarioId, variants) {
   const modificationIdPairs = new Map()
   const timetableIdPairs = new Map()
+
+  /**
+   * Create all the new modifications and new timetables while adding their new
+   * IDs to Maps that allow you to loop through the modifications a second time
+   * and correctly pair the newly created item's fields (ex: phaseFromTimetable)
+   */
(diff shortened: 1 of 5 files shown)


lib/utils/copy-modifications.js
```diff
@@ -0,0 +1,90 @@
+import uuid from 'uuid'
+
+import {ADD_TRIP_PATTERN, CONVERT_TO_FREQUENCY} from '../constants'
+
+export default function copyModifications (modifications, newScenarioId, variants) {
+  const modificationIdPairs = new Map()
+  const timetableIdPairs = new Map()
+  const newModifications = modifications.map((modification) => {
+    const id = uuid.v4()
+    modificationIdPairs.set(modification.id, id)
```
mattwigway: Don't we need to do one pass over the modifications before we start modifying them? I think this will only work if the `phaseFromTimetable` is in a modification that happens to be encountered before the modification containing the phasing.

trevorgerhardt: That's how this implementation is done. It creates the old modification id -> new modification id maps and the old timetable id -> new timetable id maps first, then runs through all the modifications with timetables and pairs them correctly. See the `matchTimetables` function below.

mattwigway: Oh, I see. This could probably use some comments.

trevorgerhardt:
Whoops! Was fixing a @flow error and thought I was changing `e.target.value` to `e.currentTarget.value` but it's a `Select` onChange!

Should be fixed now @mattwigway!

PR #423, opened 2017-06-07
Osm upload

mattwigway:


mattwigway:
@trevorgerhardt do we know why the uuid of the mock project changed?

mattwigway:
So it appears that it is not possible to make the tests pass because the uuid of that mock project changes every time you run the test suite.

mattwigway:
@trevorgerhardt tests are fixed! Problem was that because I was stringifying the project by hand now the whole stringified JSON was winding up in the snapshot instead of [object Object] or whatever.

__mocks__/uuid.js
```diff
@@ -0,0 +1,6 @@
+const uuid = require.requireActual('uuid')
+
+module.exports = {
+  ...uuid,
+  v4: () => '00000000-0000-0000-0000-000000000000'
```
trevorgerhardt: 👌 👌

trevorgerhardt:
I like that implementation! We've worked around the `uuid`s in the past—no need to anymore!

PR #427, opened 2017-06-08
Styling v2

trevorgerhardt:
Some major changes

* SIDEBAR!
* Shorter URLs (gets the project id from the scenario)
* Removed analysis button from Variants. Now selected while in Analysis mode
* Removed the comparison checkbox, just select the comparison to enable it and clear to disable it

Likelihood that I broke things are high. Snapshot diffs aren't going to be that useful since they'll be so massive so I'll wait to update them. Been playing around as much as possible to catch bugs here and there but I know there will be one or two small ones. Apologies in advance!

But I know you really just want to see it so...
screenshot 2017-06-08 10 44 03
screenshot 2017-06-08 10 46 58
screenshot 2017-06-08 11 27 05


mattwigway:
A few things from UI testing (haven't reviewed the code yet).
- It looks really good in general
- The "loading" spinner wobbles
- Let's replace the wrench icon for a scenario with a pencil
- Can we somehow deëmphasize/move the "Download Scenario" button from the variant selector? Now that we have performant built in analysis features it's not particularly useful.
- Everything on the left looks really slick, but the transit line map editor looks clunky in comparison - can we think about the design of that a bit?
- This text does not meet [web contrast guidelines](http://webaim.org/resources/contrastchecker/), and in fact I can barely read it - we need to change the colors to provide more contrast. (The white on green for upload indicator and new regional analysis also don't, and moreover the green does not match between these two buttons)
image

mattwigway:
A few more things I've noticed in testing:
- When you create a new scenario, you are not taken to the scenario.
- It seems editing modifications and then returning to analysis mode still sends the old modifications (selector issue?)
- Given the new interface style I think we want to preserve all analysis setting when entering analysis mode (i.e. not reset the variant selector to variant 0)

mattwigway:
One more thing - looks like deleting a scenario sends you to an application error (the scenario does get deleted though, I think it's trying to read the project ID from a scenario that no longer exists).

trevorgerhardt:
@mattwigway latest is up on staging 👌

configurations/default/messages.yml
```diff
@@ -106,15 +106,16 @@ analysis:
...
-  createGrid: Create a new grid
+  createGrid: Upload new opportunity data
   indicator: Indicator
   cutoff: Travel time cutoff (minutes)
-  addGrid: Add grid
+  addGrid: Add opportunity data
   uploading: uploading...
-  gridName: Grid name
-  gridFiles: Grid files
+  gridName: Opportunity data name
```
mattwigway: I thought we decided on "opportunity dataset"

lib/components/analysis/index.js
```diff
@@ -15,75 +12,106 @@ import Regional from './regional-analysis-selector'
...
+  scenarioApplicationErrors?: Array<{}>,
+  scenarioApplicationWarnings?: Array<{}>,
+  scenarioId: string,
+  scenarioName: string,
+  scenarios: Array<{
+    id: string,
+    name: string,
+    variants: string[]
+  }>,
+  percentileCurves?: Array<{}>,
```
mattwigway: Is there a preference for Array<{}> over Object[]?

lib/components/analysis/index.js
```diff
@@ -136,46 +164,54 @@ export default class SinglePointAnalysis extends Component {
...
-          [east, south],
-          [west, south],
-          [west, north]
-        ]]
-      },
-      request: {
-        ...profileRequest,
-        scenario: {
-          modifications: modifications.map(convertToR5Modification)
+    if (regionalAnalysisBounds) {
```
mattwigway: How can regionalAnalysisBounds be null/undefined?

lib/components/analysis/index.js
```diff
@@ -195,36 +231,38 @@ export default class SinglePointAnalysis extends Component {
...
-        modifications: comparisonModifications,
-        bundleId: comparisonBundleId
+        scenarioId,
+        variantIndex,
+        modifications,
+        bundleId
       }
 
-    fetchTravelTimeSurface({ scenario, comparison, bounds: projectBounds })
+      const comparison = !comparisonInProgress ? null
```
mattwigway: Not in this pull, but there is a lot of duplicated code here (see the bookmark chooser and isochrone map component)

lib/components/analysis/index.js
```diff
@@ -235,24 +273,30 @@ export default class SinglePointAnalysis extends Component {
...
 
-  setRegionalAnalysisName = (e) => this.setState({regionalAnalysisName: e.target.value})
-  setRegionalAnalysisPercentile = (e) => this.setState({regionalAnalysisPercentile: e.target.value})
+  setRegionalAnalysisName = (e: Event & {target: HTMLInputElement}) =>
+    this.setState({regionalAnalysisName: e.target.value})
+  setRegionalAnalysisPercentile = (e: Event & {target: HTMLInputElement}) =>
+    this.setState({regionalAnalysisPercentile: Number(e.target.value)})
 
-  setRegionalAnalysisBounds = (e) => {
+  setRegionalAnalysisBounds = (e: {value: string}) => {
```
mattwigway: I don't think the value should be a string, shouldn't it be a BoundingBox?

lib/components/analysis/index.js
```diff
@@ -342,88 +394,89 @@ export default class SinglePointAnalysis extends Component {
...
+          {this.renderComparisonSelect()}
 
           <Group id='select-analysis-indicator'>
             <Select
               id='select-analysis-indicator'
               name={messages.analysis.indicator}
               options={indicators.map(({ name, key }) => { return { value: key, label: name } })}
               onChange={this.changeIndicator}
-              placeholder='Select a grid...'
+              placeholder='Select an existing indicator...'
```
mattwigway: This needs to move to messages.

lib/components/analysis/stacked-percentile-selector.js
```diff
@@ -104,7 +104,7 @@ export default class StackedPercentileSelector extends Component {
       percentileCurves={percentileCurves}
       comparisonPercentileCurves={comparisonPercentileCurves}
       width={600}
-      height={400}
+      height={300}
```
mattwigway: I probably  did this but these should be pulled out to symbolic constants.

lib/components/analysis/stacked-percentile.js
```diff
@@ -198,36 +197,13 @@ export default class StackedPercentile extends DeepEqualComponent<void, Props, v
...
-  }
-
   // TODO put this in its own DeepEqualComponent so it is not re-rendered when the cutoff changes
   /** boundaries are the boundaries between slices, as array indices in percentileCurves */
   renderSlices ({ percentileCurves, color, breaks }: {
     percentileCurves: number[][],
     color: any,
     breaks: number[]
-  }): Element {
+  }) {
```
mattwigway: Do we not need the return type of these functions?

lib/components/edit-project.js
```diff
@@ -31,17 +29,27 @@ type Props = {
...
+type State = Props & {saving: boolean}
+
+export default class EditProject extends Component<void, Props, State> {
+  _created: boolean
+  _hasBeenDeleted: boolean
+
   state = {...this.props, saving: false}
 
   componentDidMount () {
+    if (!this.props.allVersions) {
```
mattwigway: Why does it only load versions if allVersions is false?

lib/components/modification/adjust-speed.js
```diff
@@ -97,8 +97,8 @@ export default class AdjustSpeed extends Component<void, Props, void> {
         }
 
         <Number
-          label='Scale'
-          name='Scale'
+          label='Scale speed by'
```
mattwigway: This needs to move to messages.

lib/components/modification/editor.js
```diff
@@ -109,7 +113,24 @@ export default class ModificationEditor extends Component<void, Props, void> {
...
       <div className='ModificationDock'>
-        <div className='ModificationDockTitle'>{name}</div>
+        <div className='ModificationDockTitle'>
+          {name}
+
+          <a
+            className='pull-right'
+            onClick={this._remove}
+            tabIndex={0}
+            title='Delete Modification'
```
mattwigway: Move to messages.

lib/components/modification/editor.js
```diff
@@ -109,7 +113,24 @@ export default class ModificationEditor extends Component<void, Props, void> {
...
+            ><Icon type='trash' />
+          </a>
+          <a
+            className='pull-right'
+            onClick={copyModification}
+            tabIndex={0}
+            title='Copy Modification'
```
mattwigway: Move to messages.

lib/components/project.js
```diff
@@ -1,59 +1,50 @@
...
+          {r5VersionUnsupported && <div className='alert alert-danger' role='alert'>
+            {messages.project.versionOutdated} <Link to={`/projects/${project.id}/edit`}>{messages.project.editProject}</Link>
+          </div>}
 
-        {isLoaded && children}
+          {children}
+        </div>
       </div>
-    )
+      : <PanelBody>Loading project...</PanelBody>
```
mattwigway: Needs to move to messages.

lib/components/scenario.js
```diff
@@ -1,80 +1,77 @@
...
-            ><span> {name}</span>
-          </Link>
+          className='ApplicationDockTitle'
+          ><Icon type='cube' /> {scenario.name}
           <Link
             className='pull-right'
-            title='Edit scenario'
-            to={`/projects/${projectId}/scenarios/${id}/edit`}
-            ><Icon type='pencil' />
+            title='Edit scenario name'
```
mattwigway: These all need to be in messages.

lib/components/select-project.js
```diff
@@ -1,33 +1,52 @@
...
-        </Button>
+        {projects.length > 0 &&
+          <div>
+            <p style={{textAlign: 'center'}}>or go to an existing project</p>
+            {projects.map((project) =>
+              <Link
+                className='BlockLink'
+                key={project.id}
+                to={`/projects/${project.id}`}
+                title='Go to Project'
```
mattwigway: Needs to go in messages.

lib/components/select-project.js
```diff
@@ -1,33 +1,52 @@
...
         </Group>
-        <Button
-          block
-          onClick={() => push(`/projects/create`)}
-          style='success'
-          >Create a new project
-        </Button>
+        {projects.length > 0 &&
+          <div>
+            <p style={{textAlign: 'center'}}>or go to an existing project</p>
```
mattwigway: Needs to go in messages.

lib/components/select-project.js
```diff
@@ -1,33 +1,52 @@
...
-          <Select
-            options={projects.map((project) => { return {value: project.id, label: project.name} })}
-            onChange={(option) => push(`/projects/${option.value}`)}
-            placeholder='Select a project...'
-            />
+          <Button
+            block
+            onClick={() => push(`/projects/create`)}
+            style='success'
+            ><Icon type='cubes' /> Create a new Project
```
mattwigway: Needs to go in messages.

lib/components/select-project.js
```diff
@@ -1,33 +1,52 @@
...
+    this.props.loadAllProjects()
   }
 
   render () {
     const {projects, push} = this.props
     return (
       <Body>
+        <div className='WelcomeTitle'>
+          <p style={{textAlign: 'center'}}>welcome to</p>
+          <span className='logo' to='/'>conveyal analysis </span>
```
mattwigway: Needs to go in messages.

lib/components/select-scenario.js
```diff
@@ -1,34 +1,61 @@
...
         <Group>
-          <Select
-            options={scenarios.map((scenario) => { return {value: scenario.id, label: scenario.name} })}
-            onChange={(option) => push(`/projects/${projectId}/scenarios/${option.value}`)}
-            placeholder='Select a scenario...'
-            />
+          <Link
+            className={`${btn}-success`}
+            to={`/projects/${projectId}/scenarios/create`}
+            title='Create a new Scenario'
```
mattwigway: All these need to go in messages.

lib/components/select-scenario.js
```diff
@@ -1,34 +1,61 @@
...
-        <Button
-          block
-          onClick={() => push(`/projects/${projectId}/scenarios/create`)}
-          style='success'
-          >Create a new scenario
-        </Button>
+
+        {scenarios.length > 0 &&
+          <div>
+            <p className='center'>or go to an existing scenario</p>
```
mattwigway: Needs to go in messages.

lib/components/sidebar.js
```diff
@@ -0,0 +1,184 @@
...
+        {outstandingRequests > 0
+          ? <div className='Sidebar-spinner'><Icon type='spinner' className='fa-spin' /></div>
+          : <Link title='Projects' to='/' className='Sidebar-logo' />}
+
+        {projectId &&
+          <div>
+            <SidebarNavItem
+              active={isProjectPath(currentPath)}
+              icon='cubes'
+              text='Scenarios'
```
mattwigway: All these need to go in messages.

lib/containers/logout.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+import {Component} from 'react'
+import {connect} from 'react-redux'
+
+import {logout} from '../actions'
+
+class Logout extends Component {
```
mattwigway: Is this a component that logs the user out when mounted? It could use a comment explaining the rationale for this.

lib/reducers/scenario.js
```diff
@@ -130,16 +131,25 @@ export const reducers = {
...
-    return {
-      ...state,
-      currentBundle: state.bundlesById[currentScenario.bundleId],
-      currentScenario,
-      scenarios: Object.values(scenariosById),
-      scenariosById,
-      variants: currentScenario.variants || []
+    if (action.payload && action.payload.id) {
+      const currentScenario = {...action.payload}
+      const scenariosById = {...state.scenariosById}
```
mattwigway: I thought these were done with selectors now?

lib/routes.js
```diff
@@ -44,19 +45,22 @@ export default function Routes ({history, store}: {
...
+      </Route>
+      <Route path='scenarios/:scenarioId' component={Scenario}>
+        <IndexRoute component={Modifications} />
+        <Route path='edit' component={EditScenario} />
+        <Route path='import-modifications' component={ImportModifications} />
+        <Route path='import-shapefile' component={ImportShapefile} />
+        <Route path='modifications/:modificationId' component={Modifications} />
+        <Route path='analysis' component={SinglePointAnalysis} />
+        <Route path='analysis/:variantId' component={SinglePointAnalysis} />
+        {/* TODO should be nested under analysis */}
```
mattwigway: Remove this comment.

lib/components/analysis/index.js
trevorgerhardt: I would start with `Array<{}>` because you could further define the object properties inline if it's not necessary to pull out into a reusable type.

lib/components/analysis/index.js
trevorgerhardt: When I converted to Flow properties I must have added a `?` to the property by accident causing a flow error when attempting to access it before checking it's existence.

lib/components/analysis/index.js
trevorgerhardt: 👌

lib/components/analysis/index.js
trevorgerhardt: It was being used below to compare to `__PROJECT` and `r.id`.

lib/components/analysis/stacked-percentile.js
trevorgerhardt: Flow can infer a lot. Especially with regards to React. `render` functions generally do not need a return type. Do you have `flow-ide` installed? This one gave me a flow error so I just removed it.

lib/components/edit-project.js
trevorgerhardt: If `allVersions` have not been loaded then load them. If they have been loaded, no need to load them again.

lib/containers/logout.js
trevorgerhardt: Yup, allows us to use it for a `react-router` route.

lib/reducers/scenario.js
trevorgerhardt: Some of this could move into selectors. But we do still have to set up the state for selectors to use.

trevorgerhardt:
@mattwigway all the text indicated has been moved into messages.yml. I'm starting to think there's a better way to do default text inline...anyway this should be good to go now 👌

trevorgerhardt:
@mattwigway now we should be good to go 👍

mattwigway:
@trevorgerhardt reports don't work with this branch, I just get a blank screen.

mattwigway:
Also, still seeing the issue mentioned above where creating a project doesn't work (r5 versions don't load) until you have opened a scenario.

mattwigway:
Also, when you have a lot of projects (maybe also scenarios) the project list does not scroll.

mattwigway:
Looks like the "out of date backend" warnings are also broken - when I have a project set to v2.2.0 I don't see a bright red warning even though that version has been pulled from S3.

mattwigway:
Also, when I select a comparison scenario and the baseline variant, the stacked percentile selector shows the name of the variant as `undefined`:

image

trevorgerhardt:
@mattwigway reporting shouldn't have changed all that much. I noticed an error with a modification similar to the errors when running analysis (no route selected, not enough stops/segments, etc.) and that caused a blank screen for me.

I added a "try-catch-render" component wrapper that we *may* want to start using in development.

PR #430, opened 2017-06-12
Add GTFS bundle preview to map while editing scenario

trevorgerhardt:
Possible future additions:

* Button to the new UI to turn on or off at any time
* Text summary of bundle, possibly searchable
* Tooltips to lines + stops including feed + id + name


trevorgerhardt:
Closing because this implementation is wayyy too slow.

PR #431, opened 2017-06-12
Report improvements

mattwigway:
* Fix printing on chrome by making bootstrap floated elements not floated.
* Add phasing information to reports.
* Add numbers to each modification in reports (basically a page number).

### DIFF
--- a/lib/components/report/phase.js
+++ b/lib/components/report/phase.js
@@ -1,0 +1,44 @@
+// @flow
+
+/** display the phasing information for a timetable entry in the report */
+import React from 'react'
+import DeepEqualComponent from '../deep-equal'
+import {sprintf} from 'sprintf-js'
+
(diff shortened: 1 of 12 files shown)


lib/components/report/phase.js
```diff
@@ -0,0 +1,44 @@
...
+} from '../../utils/timetable'
+
+type Props = {
+  timetable: Timetable,
+  scenarioTimetables: Timetable[],
+  feedScopedStops: GTFSStop[]
+}
+
+export default class Phase extends DeepEqualComponent<void, Props, void> {
+  render (): HTMLElement {
```
trevorgerhardt: The return type on render can be inferred by flow and isn't really needed

mattwigway: I think I did this because it was complaining, but that makes sense - because it's defined in Component, right?

PR #435, opened 2017-06-20
Allow selecting multiple routes in Adjust Speed modifications

trevorgerhardt:


lib/actions/get-feeds-routes-and-stops.js
```diff
@@ -154,14 +154,15 @@ function getUniqueRouteIdsFromModifications (modifications) { // TODO: this can
       modification.type === REROUTE) {
       // NB we are in fact getting all routes in every feed that have this route ID, which means we're pulling down a bit more data than we need to but it's pretty benign
       if (modification.routes !== null) {
-        modification.routes.forEach((r) => {
-          if (routeIds.indexOf(r) === -1) {
-            routeIds.push(r)
-          }
-        })
+        // TODO This needs to be optimized on the server side
+        const routeId = modification.routes[0]
```
mattwigway: Is the idea here that we don't actually need detailed information about all routes in a modification with multiple routes?

mattwigway: Because if so this needs a comment about why it's acceptable to just use the zeroeth element.

mattwigway:
Other than that :+1: LGTM.

mattwigway:
Actually, I guess I should test it _in situ_ and make sure it actually has the intended effect before I approve.

mattwigway:
So this does appear to be working, however in most of the feeds I tried it on the scenarios failed to apply because there are routes with no trips (which are represented in the frontend but are dropped by R5). We need to come up with a way to work around that but that could be a separate pull in the interests of merging outstanding branches.

PR #450, opened 2017-07-13
Sampling distributions

### LINKED ISSUE #436: Show sampling distribution on click in regional analysis results
In the same vein as #434, but for the regional results. I think we should display it as a KDE rather than an boxplot just to avoid confusing the user by using two very similar plots for two very different things.

This will require rendering the access grids to gridualizer on the client rather than the server, or making something on the server that will sample from the access grids. This might overwhelm the memory in the browser (a single access grid contains up to 1000 bootstrap replications at each pixel, vs. 5 percentiles in travel time surfaces).

mattwigway:
Add sampling distribution display on click to regional analyses, fixes #436. Depends on sampling-distributions branch of analysis-backend, should not be merged until that is.


trevorgerhardt:
Let me know when this is ready to take a look 👌

PR #452, opened 2017-07-14
Filter out deleted analyses in regional comparison selector

landonreed:


trevorgerhardt:
@mattwigway @landonreed should we even be returning deleted analyses to the UI?

landonreed:
Probably not.  Maybe we should add a flag to the controller that will only return deleted objects if you ask for them `deleted=true`.

PR #453, opened 2017-07-18
Weighted averaging

mattwigway:
This depends on an R5 change (the `pixel-weights` branch) so shouldn't be merged until that is merged, but it's ready for review.

I fixed almost all the flow errors but there's still one about the State type not being defined that I can't figure out.


mattwigway:
This should be done, there are a flow errors I can't figure out how to fix, and also a snapshot test error that I'm also seeing in dev.

PR #498, opened 2017-09-12
Help tooltips

ansoncfit:
Uses react-tooltip to add instructions about opportunity dataset and GTFS bundle upload.

Adds a ❔ to those panels.  When users hover over ❔, a tooltip with abbreviated instructions appears.  We could consider making the ❔s link to the appropriate readthedocs pages, but the maintenance required to update links as readthedocs changes might not be worth it.

### DIFF
--- a/lib/components/edit-bundle.js
+++ b/lib/components/edit-bundle.js
@@ -2,6 +2,7 @@ import Icon from '@conveyal/woonerf/components/icon'
 import Pure from '@conveyal/woonerf/components/pure'
 import PropTypes from 'prop-types'
 import React from 'react'
+import ReactTooltip from 'react-tooltip'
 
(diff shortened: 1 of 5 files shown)


lib/components/edit-bundle.js
```diff
@@ -114,6 +115,12 @@ export default class EditBundle extends Pure {
       <Panel>
         <Heading>
           {bundleId ? messages.bundle.edit : messages.bundle.create}
+          <a data-tip='React-tooltip'>
+          <Icon className='fa-question-circle-o fa-lg pull-right' / >
```
trevorgerhardt: Icons accept a type so you can do `<Icon type='question-circle-o' className='pull-right' />`

PR #504, opened 2017-09-13
Explanatory tooltips (now on latest base branch)

ansoncfit:
Uses react-tooltip to add instructions about opportunity dataset and GTFS bundle upload.

Adds a ❔ to those panels. When users hover over ❔, a tooltip with abbreviated instructions appears. We could consider making the ❔s link to the appropriate readthedocs pages, but the maintenance required to update links as readthedocs changes might not be worth it.

### DIFF
--- a/lib/components/create-grid.js
+++ b/lib/components/create-grid.js
@@ -29,12 +29,10 @@ export default class CreateGrid extends Component {
       <Panel>
         <Heading>
           {messages.analysis.createGrid}
-          <a data-tip='React-tooltip'>
-            <Icon className='fa-question-circle-o fa-lg pull-right' / >
+          <a data-tip={messages.analysis.createGridTooltip}>
+            <Icon type='question-circle-o' className='fa-lg pull-right' />
+            <ReactTooltip className='help-tooltip' place='right' />
           </a>
-          <ReactTooltip className='help-tooltip' place='right'>
-            {messages.analysis.createGridTooltip}
-          </ReactTooltip>
         </Heading>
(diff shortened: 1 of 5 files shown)


lib/components/create-grid.js
```diff
@@ -28,6 +29,12 @@ export default class CreateGrid extends Component {
       <Panel>
         <Heading>
           {messages.analysis.createGrid}
+          <a data-tip='React-tooltip'>
+            <Icon className='fa-question-circle-o fa-lg pull-right' / >
```
trevorgerhardt: For the most part `Icon`'s will just have a type. So let's go with `<Icon type='question-circle-o' className='pull-right' />` in these instances for consistency.

lib/components/create-grid.js
```diff
@@ -28,6 +29,12 @@ export default class CreateGrid extends Component {
...
+          </a>
+          <ReactTooltip className='help-tooltip' place='right'>
```
trevorgerhardt: Does this need to be *next* to the component associated with it or can it be inside it? If it is self-contained that would be nice.

trevorgerhardt:
@ansoncfit don't worry about the tests breaking for now.

trevorgerhardt:
Also, the order goes

* Owner submits a PR, asks for review
* Reviewer approves
* Owner merges

So feel free to merge!

lib/components/create-grid.js
ansoncfit: Oops, thanks for the extra reminder.

PR #517, opened 2017-10-03
Rename hierarchy

ansoncfit:
User feedback has convinced us that "variant" is a confusing term.

We are moving toward a new hierarchy, and this PR mostly implements an intermediate set of changes in the UI, renaming variants to scenarios, scenarios to projects, and projects to regions.

This leaves us in the unfortunate position of having the user-facing nomenclature not match our file names.  But we'll be eliminating some level of the hierarchy soon, and that will be a good chance to rename files etc.

This PR also includes a start at associated changes to the user manual, and a couple miscellaneous fixes for readability/legibility/consistency.

### DIFF
--- a/configurations/default/messages.yml
+++ b/configurations/default/messages.yml
@@ -73,7 +73,7 @@ project:
   createActionTooltip: Each project can include scenarios with modifications that are applied to the same GTFS bundle.
   delete: Delete project
   deleteConfirmation: Are you sure you would like to delete this project?
-  editAction: Save changes to project
+  editAction: Save project
   editTitle: Edit project
   editName: Edit project name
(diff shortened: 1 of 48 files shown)


trevorgerhardt:
@ansoncfit nice! Will take a look ASAP. Have you run `mastarm lint-messages` to see if there were any missing messages with such a big change?

trevorgerhardt:
@ansoncfit it appears that there are two messages missing when `mastarm lint-messages` is run, a few JavaScript linter errors that need to be fixed when `mastarm lint` is run, and snapshots that need to be updated when `mastarm test` is run. Let me know if you need help with any of that!

PR #518, opened 2017-10-04
Simplify edit bundle and modify it to edit feed names

trevorgerhardt:
This fixes the currently broken "edit bundle" form. Initial implementation took a different approach to display more information about the feed while editing.


abyrd:
These codecov comments are huge and end up being the first thing I see in my email / on Github for every commit.

trevorgerhardt:
Disabled for now. Will create a subsequent pull request to fix the issues described above.

PR #521, opened 2017-10-05
Add front-end options to set access/egress parameters

ansoncfit:
Users should be able to select between walking/bike for access/egress legs, and set speeds for these modes.  This PR adds related components to the front-end, but disables some of them until back-end issues are resolved.

### DIFF
--- a/lib/components/analysis/egress-mode-selector.js
+++ b/lib/components/analysis/egress-mode-selector.js
@@ -51,6 +51,7 @@ export default class EgressModeSelector extends PureComponent {
           </Button>
           <Button
             active={nonTransitMode === BICYCLE}
+            disabled
             onClick={this._selectEgressMode(BICYCLE)}
             title={messages.analysis.modes.bicycle}
           >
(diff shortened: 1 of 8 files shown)


lib/components/analysis/egress-mode-selector.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+const WALK = 'WALK'
+const BICYCLE = 'BICYCLE'
+
+/** Select modes of travel */
+export default class EgressModeSelector extends PureComponent {
+  props: Props
+
+  _selectEgressMode = memoize(newMode => () => {
```
evansiroky: What is the purpose of memoizing this method?

evansiroky: Oh, I see how this is used.  Nice pattern, I'm going to start using this pattern in my codebases.

lib/components/analysis/mode-selector.js
```diff
@@ -90,7 +91,7 @@ export default class ModeSelector extends PureComponent {
           </Button>
           <Button
             active={!!transit}
-            onClick={this._toggleTransit}
+            // uncomment to allow transit toggling onClick={this._toggleTransit}
```
trevorgerhardt: We should disable this button than, correct?

PR #525, opened 2017-10-12
Fix and enable bundle editing

trevorgerhardt:
Fix mismatches when editing bundles and navigating to projects afterwards.

NOTE: This does not modify/fix how we name feeds (from the agency list) aka the Amtrak -> Martz Trailways problem that @ansoncfit noticed. I initially went down the path of updating that but due to discussions of possibly changing how we handle bundles/feeds in the future I didn't end up wanting to muck around with that too much. 

So you can just manually change it to the appropriate name for now 👍


ansoncfit:
Should the edited feed name show up when creating modifications?

Saved changes to bundle:
image

But new feed name is not reflected in pre-existing modification or in a newly created modification:
image

Issue #510, opened 2017-09-20
Allow management of opportunity datasets


ansoncfit:
Multiple users have requested better control/management of opportunity data.  I imagine the Opportunities Dataset panel looking like:

-Pointset/shapefile 1 [rename | use as sub-area | remove from list of sub-areas]
--                  [use all as grids | remove all from list of grids]
--Attribute 1 [rename | use as grid | remove from list of grids]
--Attribute 2 [rename | use as grid | remove from list of grids]

-Pointset/shapefile 2 [use as sub-area | remove from list of sub-areas]
--                  [use all as grids | remove all from list of grids]
--Attribute 1 [rename | use as grid | remove from list of grids]

-LODES (if available)
--                  [use all as grids | remove all from list of grids]
--Attribute 1 [rename | use as grid | remove from list of grids]
--Attribute 2 [rename | use as grid | remove from list of grids]
etc.

This would help avoid dropdowns that are bloated and hard to search through for destination opportunities (and origin weighting classes when viewing regional results).  I would even advocate having the default be for attributes not to show up in lists of grids.  Most users will only want a few of the available LODES attributes.  And even when uploading their own shapefiles, users often include extraneous columns like area or population density.

This could also address https://github.com/conveyal/analysis-ui/issues/509

ansoncfit:
See also https://github.com/conveyal/analysis-ui/issues/266

abyrd:
Also, if someone uploads a particular shapefile, then uploads a newer/updated version of the same shapefile, we'll get a new set of grids with exactly the same names. It's then impossible to know which ones came from which input file. A grid management page with creation date and editable names etc. would help a lot here.

PR #528, opened 2017-10-24
Refactor Grids/Indicators to be Opportunity Datasets

trevorgerhardt:
Add the ability to view and remove datasets.

Create a module to encapsulate everything related to Opportunity Datasets and the management of
them. This is also a prototype for how to build modules going forward. Required a lot of name
changes and re-arrangements of files and tests.

addresses #510 

depends on https://github.com/conveyal/analysis-backend/pull/67 to work


lib/modules/opportunity-datasets/actions.js
```diff
@@ -0,0 +1,160 @@
...
+    }))
+  }
+}
+
+export const checkUploadStatus = (handle: string, projectId: string): FetchAction =>
+  fetch({
+    url: `/api/grid/status/${handle}`,
+    next (error, response) {
+      if (!error) {
+        if (response.value.status === 'DONE') {
```
ansoncfit: Should this be response.statusText?

image

PR #529, opened 2017-10-24
Added ability to turn on/off bus, tram, subway, and rail

### LINKED ISSUE #489: Mode selector
We need to be able to select transit modes (at least activate / deactivate rail, bus).

trevorgerhardt:
Separates out the mode selection for transit into four types. You must have at least one selected at
all times at the moment due to an error in r5.

closes #489


lib/components/analysis/mode-selector.js
```diff
@@ -35,70 +42,105 @@ export default class ModeSelector extends PureComponent {
...
+  }
+
+  _toggleTransitMode = memoize(mode => () => {
     const {profileRequest, setProfileRequest} = this.props
+    let transitModes = this._hasTransit(mode)
+      ? without(profileRequest.transitModes.split(','), mode).join(',')
+      : `${profileRequest.transitModes},${mode}`
+
+    if (transitModes === '') {
+      window.alert('At this time you must have at least one transit mode turned on.')
```
ansoncfit: Maybe change to "For the moment, you must have at least one transit mode selected"

Otherwise, people might read that as relating to the time window they specified for analysis.

trevorgerhardt: Good call 👌

PR #536, opened 2017-11-01
Region selection fix, tooltips, and background changes

### LINKED ISSUE #349: Regional bounds selector not exactly matching bounds
In the Atlanta project on staging, in the Low Frequency High Speed scenario, creating a new regional analysis with the same size as an existing one with a single origin results in a two-origin scenario. I suspect this is a boundary condition due to the 1x1 dimensions of the source analysis.

### LINKED ISSUE #514: Inform users of minimum allowable stop spacing
Either with a tooltip that shows up when they try to enter an invalid number, or by re-writing to the minimum or DEFAULT_STOP_SPACING_METERS.

A 50 meter limit seems hard coded into https://github.com/conveyal/analysis-ui/blob/dev/lib/components/modification/edit-alignment.js#L124

### LINKED ISSUE #530: "Create a bundle" button not active
On Chrome, the Create a bundle button doesn't do anything (link : analysis.conveyal.com/#)

### LINKED ISSUE #532: Create new region "pre-filled" with existing region details
When we created a new region for SBB, the form to fill in the region details (bounds name etc.) was already pre-filled with the details of the existing region. We backed out and tried again to create a new region; the second time the fields were all empty in the form so we proceeded to create the new region.

### LINKED ISSUE #533: Cannot manually enter region bounds
We wanted to use some specific bounds recorded from another region, but it's impossible to type them into the fields when creating a new region.

### LINKED ISSUE #534: Loading of/switching between regional results layers
When loading of / switching between regional results layers, the url of the results layer is properly generated.

However I need to manually reload the page in the browser (Version 61.0.3163.100 (Official Build) (64-bit) on OS X 10.10.5) to display the results.

### LINKED ISSUE #535: Changing regions does not change list of projects
An extra page refresh fixes.

First region:
image

After changing to a second region:
image

After refreshing:
image

trevorgerhardt:
I ended up mixing in a few bug fixes with minor enhancements on the same branch.

This PR
- Moves the editing dock back into the main dock
- Fixes #535 
- Changes how we "select" active projects and modifications
- Allows clearing of feeds from modifications in cases where the feeds no longer exist. It shouldn't happen, but happens in testing enough that it seemed like the chute was worth it to add
- Fixes #530 by reducing confusing by removing the "Create a bundle" link button in the heading that doesn't correlate to form itself 
- Always show the active modification on the map
- Adds a note about the minimum stop spacing, fixes #514
- Show correct empty form when creating a region, fixes #532 
- Allow manually editing the region bounds, fixes #533 
- Fix switching between regional analyses, fixes #534 
- Fix off by one error causing a mismatch when comparing regional bounds, fixes #349


ansoncfit:
Should deleting a modification return the user automatically to the main editing dock/list of modifcations, instead of staying on that modification, which just say "loading"?

image

ansoncfit:
While you're working on this, could you remove the title name field, and just show it in the title block if the user clicks on a rename icon (for the moment, I guess we could keep overusing ✏️ ).  As we're currently forcing users to enter a name in the add modification modal dialog, having an editable field show up by default seems duplicative/confusing.

image

lib/components/bundles.js
```diff
@@ -40,23 +40,26 @@ export default class Bundles extends Component {
...
         </div>
         <div className='InnerDock'>
           <div className='block'>
             <br />
 
             <p>Bundles are a collection of one or more GTFS feeds.</p>
 
-            <Group label='Bundles'>
+            <Group>
+              <Button
```
ansoncfit: Can this button be hidden if there are no existing bundles?

https://github.com/conveyal/analysis-ui/issues/530

ansoncfit:
If a modification is hidden from the map ( 👁 has a slash through it), it won't show up, even when the modification is being actively edited.  I would suggest that it be shown when being actively edited, regardless of the toggle map display setting.

ansoncfit:
I haven't been able to get stops for one route to show up in adjust dwell times or remove stops modifications.

lib/components/bundles.js
trevorgerhardt: Ah, meant to do that originally. Now correctly addresses #530 by never showing two create buttons.

ansoncfit:
To clarify the inability to get stops to show up, it seems like the pattern selector is not loading for adjust dwell time, remove stops, and reroute modifications until the page is refreshed.

PR #547, opened 2017-11-06
Model refactoring

### LINKED ISSUE #542: Multi-user locking
Projects should not be able to be modified by more than one user at a time to prevent data loss and unintentional behavior.

### LINKED ISSUE #550: Drop-down of bundles is cut off in create project page
image

### LINKED ISSUE #568: List of projects cut off
image

trevorgerhardt:
This PR was created to close #542 and support corresponding changes made to the backend (https://github.com/conveyal/analysis-backend/pull/72) and include a few extra things that I refactored/fixed along the way.

- Properly updates each model on change so that it has the correct `nonce`.
- Changes all `id` -> `_id` to properly mimic how they are represented in the database (_Note: A very high percentage of the non-snapshot changes were these_)
- Moves all model `_id` creation to the server
- Changes `scenario` to `scenarioId` in modifications
- Moves copying a modification and importing all modifications from another project to the server side
- Adds `@flow` support and corresponding types to as many of the touched files as reasonable
- Creates selectors for common state selection
- Only retrieve the opportunity dataset when doing analysis or viewing it
- Switches places we used `<Panel>` to use `<InnerDock>` (fixes #550 and fixes #568)
- Adds each user's email address to the encrypted JSON web token so we can track who created and edited each object.
- And more...


lib/actions/project.js
```diff
@@ -2,7 +2,6 @@
 import fetch from '@conveyal/woonerf/fetch'
 import {createAction} from 'redux-actions'
 import {push} from 'react-router-redux'
-import uuid from 'uuid'
```
ansoncfit: This is what you mean by "Moves all model _id creation to the server" in the PR comment?

trevorgerhardt: Yup!

trevorgerhardt: Here: https://github.com/conveyal/analysis-backend/blob/8d5c4068e653d62ad50b542ee873450d810f1649/src/main/java/com/conveyal/taui/persistence/MongoMap.java#L107

lib/actions/scenario.js
```diff
@@ -77,91 +82,88 @@ export const saveToServer = scenario =>
...
-export const setAndLoadModifications = scenario => [
-  set(scenario),
-  getModificationsForScenario({
-    bundleId: scenario.bundleId,
-    scenarioId: scenario.id
-  })
-]
-
 /**
  * Scenario Variants
```
ansoncfit: If we have a comment here, it might be worth noting that "variants" are called "scenarios" in the interface.

[my fault for pushing mismatching nomenclature] 😅

lib/components/application.js
```diff
@@ -235,16 +238,21 @@ function UiLock ({clearError, error, errorMessage}) {
   return (
     <Modal onRequestClose={clearError}>
       <ModalTitle>
-        Oh no! Looks like an error has occured.
+        Unfortunately it appears an error has occured...
```
ansoncfit: As long as we're editing this message, could it be moved to messages.yml? (or create a TODO here, and I can follow up).

lib/components/edit-scenario.js
```diff
@@ -82,27 +81,29 @@ export default class EditScenario extends PureComponent {
...
+                    options={bundles.map(bundle => {
+                      return {value: bundle._id, label: bundle.name}
+                    })}
+                    onChange={option =>
+                      this.setState({...this.state, bundleId: option.value})}
+                    placeholder='Select a GTFS bundle...'
+                    value={bundleId}
+                  />
+                </FormGroup>
+              : <p>There are no GTFS bundles for this region! You must create one before you can a project.</p>}
```
ansoncfit: I think you meant "..before you can **create** a project."

More generally, the first part of this could feel a bit accusatory.  Maybe "To get started with projects in this region, first create a GTFS bundle"

lib/components/application.js
trevorgerhardt: Sure sure!

lib/components/modification/segment-speeds.js
```diff
@@ -46,7 +47,7 @@ export default class SegmentSpeeds extends PureComponent<void, Props, State> {
   }
 
   componentWillReceiveProps (nextProps: Props) {
-    if (nextProps.segmentDistances !== this.props.segmentDistances) {
```
ansoncfit: Just curious -- does this change functionality, or is it a stylistic choice?

lib/components/report/index.js
```diff
@@ -62,39 +64,10 @@ export default class Report extends Component<void, Props, void> {
       feedScopedStops
     } = this.props
 
-    if (modifications == null || scenario == null || bundle == null) {
+    if (!modifications || !scenario || !bundle) {
       return <div />
     }
 
-    // ensure all feeds are loaded
```
ansoncfit: Nice, glad this block is being retired.

lib/components/modification/segment-speeds.js
trevorgerhardt: This is to prevent unnecessary calculation when the exact array isn't returned but the contents are the same: `[1, 2, 3] !== [1, 2, 3]` but `isEqual([1, 2, 3], [1, 2, 3]) === true`

PR #552, opened 2017-11-09
Enable analysis biking access and analysis with no transit modes

trevorgerhardt:


abyrd:
@trevorgerhardt can you explain more fully what this PR does? I'm looking through the code and can only really discern that it adds the text "CURRENTLY DISABLED" in several places, while adding and removing `disabled={true}` or `disabled={disabled}` from several places. Based on the title, I suppose the objective is to make the biking mode available in the UI, and allow analysis to be performed when no transit modes are selected. But what does the "CURRENTLY DISABLED" text do? Is this just to inform users that certain combinations don't work yet, or work only with experimental worker versions?

trevorgerhardt:
You are correct. The note is to inform that it doesn't currently work. Leaving it in the UI without a note makes it seem like there should be a way for the user to enable it, but there is not.

PR #563, opened 2017-11-14
Move R5 Version selection into analysis

### LINKED ISSUE #539: R5 version should be set at the analysis level, not the region level
It should default to the same version on which analysis-backend depends, and the available versions in the drop-down should be restricted those matching that major version number.

In terms of warnings, the front-end could check if a graph for a given network and R5version is already built (e.g. there's a [networkId]_[selectedR5Version].dat in the graphs bucket).

If there's not one already built, show a warning on the analysis page: "A baseline network with the selected GTFS bundle and Analysis backend version has not been built yet, so initializing a compute cluster may take longer than usual."  

If there is no [networkId]_[selectedR5Version].dat but there is [networkId]_*.dat, append warning: "Results may differ slightly from past analyses, which used other Analysis backend versions."

trevorgerhardt:
Closes #539


configurations/default/messages.yml
```diff
@@ -40,6 +40,13 @@ phasing:
   headwayMismatchWarning: Warning! The selected timetable's %(selectedTimetableHeadway)s minute headway does not equal this timetable's headway, which could cause unintended behavior.
   noAvailableStopsWarning: The selected timetable does not contain stops available to phase from!
   exactTimesWarning: Exact times is not available because phasing is enabled on this timetable.
+
+r5Version:
+  analysisVersionDifferent: WARNING: You have completed regional analyses with a different r5 version which could change results slightly.
```
ansoncfit: Is the change in user-facing terminology from "Analysis backend" to "r5 version" intentional?

If we're open to changing it, I like the term "Analysis engine."  "Engine" seems to be used by other software, e.g. https://www.safe.com/fme/fme-server/

ansoncfit: Suggest "... version. Using different routing engines may change results."

configurations/default/messages.yml
```diff
@@ -40,6 +40,13 @@ phasing:
...
+  comparisonIsDifferent: WARNING: The analyses you are comparing ran on different versions of r5 which could change results slightly.
```
ansoncfit: Suggest "...versions of the analysis engine.  Changes in accessibility may be due to differences in the routing engines, rather than the transport scenarios."

lib/actions/analysis/regional.js
```diff
@@ -25,7 +27,15 @@ export const setRegionalAnalyses = createAction('set regional analyses')
 export const load = (projectId: string) =>
   fetch({
     url: `/api/project/${projectId}/regional`,
-    next: (error, response) => !error && setRegionalAnalyses(response.value)
+    next (error, response) {
+      const analyses = sortBy(response.value, a => -a.createdAt) // newest at the top
```
ansoncfit: I (and I think at least one client) sometimes use the naming of regional analyses to sort/group them intentionally.  Could we give users the options to sort either by name or createdAt?

lib/modules/r5-version/constants.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+
+// Bucket where R5 JARs are to be found
+export const R5_BUCKET = 'https://r5-builds.s3.amazonaws.com'
+
+// Recommended r5 version
```
ansoncfit: Nice.

Probably worth adding a note about this file and these settings to deployment docs.

lib/actions/analysis/regional.js
trevorgerhardt: Let's just decide on one or the other and go with that. Most recently created seemed the most logical to me and is used for selecting the default R5 version. Either way I'd sort for that at some point.

configurations/default/messages.yml
trevorgerhardt: Well, it's definitely not the "Analysis Backend" so I was just going to call it what it is but "routing engine" sounds good to me.

Issue #105, opened 2016-05-30
Cannot clear from/to stop in reroute modifications


abyrd:
When editing a reroute modification, once a from- or to-stop is selected it is impossible to clear that selection. This means that if you've made a complex reroute near the end of a route and then decide to actually move the terminus of the route, you can't.

However, I'm wondering whether the from-stop and to-stop are necessary at all. It may be possible to eliminate them from the editor and the modification messages. See #104.

ansoncfit:
I've just run into this issue and can confirm it's frustrating.  You can actually drag the stops around so it looks like it should be valid (as noted in #104):

![image](https://user-images.githubusercontent.com/2173529/31105193-3ab97c12-a7b0-11e7-888c-afb8b221046d.png)

But when trying to do a single point analysis the UI shows a "script error" without further explanation.  The error in the console states (correctly) that "First stop in reroute is not fromStop!"

A quick fix would be to add a "clear" button next to the "select" from/to stop buttons.

Another suggestion would be to add a "make start of reroute" and "make end of reroute" tooltip that appears when a user clicks on a (green) stop that's part of the existing pattern, and show a warning ("Click a green stop to connect your reroute to an existing stop") in the editing panel until at least one of these was clicked.

Issue #104, opened 2016-05-30
From and to stop fields not necessary in reroute modifications


abyrd:
When creating a reroute modification we have two steps: first the from and to stops are selected, then an alignment is specified between those two endpoints. This has confused some users of the editor. In fact we don't need two different steps. When you edit the alignment, we can just require the first and/or last click to be on an existing stop.

abyrd:
It is possible to move the first or last stop in the reroute alignment so it does not correspond with the from-stop or to-stop. This is apparently the cause of #103.

mattwigway:
This is still the case.

Nate-Wessel:
It seems like the problem is not that the first/last stop can be moved away from an existing stop, (thus creating a new stop) but that doing so doesn't issue any warnings or unselect that stop as the start/end of the detour. When transitioning out of editing mode, the user should be warned if they don't have a start/end stop selected.

Also, I notice that dragging the start/end stop away from an existing stop turns the stop from black to blue (existing -> new) _and_ that dragging it back to the original position seems to do the opposite (new -> existing, blue -> black). If that were just another mechanism for selecting/unselecting start/end stops, things might work better.

ansoncfit:
@abyrd and I just discussed the second note above -- users can move the start/end stop away from an existing stop on the pattern.

abyrd:
In discussion today, there is some hesitation over whether all use cases described in https://docs.conveyal.com/edit-scenario/modifications#reroute can be covered if start/end stop fields are eliminated. Specifically the case of truncating a line (empty reroute geometry) was mentioned. What is clear is that multiple workflows are possible, and the current system does work so it's not a high priority to rework this until we can do a proper design session.

PR #575, opened 2017-11-17
Reroute modifications

ansoncfit:
This PR attempts to make reroute modifications more intuitive for users.  

1. From Stop and To Stop selection buttons are relabeled, clarifying that these are existing stops on the currently selected pattern.  
2. Edit alignment is disabled until fromStop or toStop has been selected
3. Users can now clear fromStop and toStop, and select different ones (resolving #105)

This is different approach than suggested in #104, but it may be usable enough to close that issue.

### DIFF
--- a/lib/components/modification/add-trip-pattern.js
+++ b/lib/components/modification/add-trip-pattern.js
@@ -49,6 +49,7 @@ export default ({
       modification={modification}
       setMapState={setMapState}
       update={update}
+      disabled={false}
     />
 
     <Timetables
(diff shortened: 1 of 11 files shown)


lib/components/modification/add-trip-pattern.js
```diff
@@ -49,6 +49,7 @@ export default ({
       modification={modification}
       setMapState={setMapState}
       update={update}
+      disabled={false}
```
trevorgerhardt: For boolean props attributes, not passing it in is usually the equivalent of passing in `{false}`.

ansoncfit: That's what I assumed, but not including it seemed to generate this complaint:

image

Issue #511, opened 2017-09-26
Allow download of grids (e.g. as geotiffs)


ansoncfit:
Users requested the ability to generate their own histograms of the distribution of accessibility, rather than using the ones in the browser.  We should enable them to download grid files (e.g. from shapefiles they have uploaded) as `.geotiff`s so they can do raster analysis with both the origin population grid and the accessibility grid.

One option would be to generate the `.geotiff`s and save them to s3 automatically when creating grids, as we already do for `.png`s (https://github.com/conveyal/analysis-ui/issues/510).  A helpful method is available at https://github.com/conveyal/r5/blob/master/src/main/java/com/conveyal/r5/analyst/Grid.java#L285

Alternatively, we could do this on request.

In the UI, we could add download as an option for each grid (see https://github.com/conveyal/analysis-ui/issues/510)

The user manual could offer some suggestions on basic raster analysis (e.g. clipping the origin/accessibility grids so they have the same extents, using GIS to convert to [`.xyz`](http://www.gdal.org/frmt_xyz.html) files that can be copied into a spreadsheet, and the [Raster package](https://geoscripting-wur.github.io/IntroToRaster/) in R.)

ansoncfit:
Making the download of raw .grid files will be very useful for the MIT project.

PR #577, opened 2017-11-28
Grid downloads

ansoncfit:
UI changes to address #511.

Depends on https://github.com/conveyal/analysis-backend/pull/78

### DIFF
--- a/lib/modules/opportunity-datasets/actions.js
+++ b/lib/modules/opportunity-datasets/actions.js
@@ -91,6 +91,37 @@ export const deleteOpportunityDataset =
       }))
     }
 
+  export const downloadOpportunityDataset =
+    (dataset: OpportunityDataset, format: string): ThunkAction =>
+      (dispatch: Dispatch, getState: GetState) => {
+        const state = getState()
+        const projectId = selectCurrentProjectId(state, {})
+        const datasets = selectOpportunityDatasets(state, {})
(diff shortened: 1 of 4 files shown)


lib/modules/opportunity-datasets/actions.js
```diff
@@ -91,6 +91,35 @@ export const deleteOpportunityDataset =
...
+    (dataset: OpportunityDataset, format: string): ThunkAction =>
+      (dispatch: Dispatch, getState: GetState) => {
+        const state = getState()
+        const projectId = selectCurrentProjectId(state, {})
+        dispatch(fetch({
+          url: `/api/opportunities/${projectId}/${dataset.key}/${format}?redirect=false`,
+          next: (err, res) => {
+            if (!err) {
+              return fetch({
+                url: res.value.url,
```
trevorgerhardt: Can you just call `window.open` on this url? It should be a fully signed URL from S3

ansoncfit: I was using the same approach as below (https://github.com/conveyal/analysis-ui/pull/577/files/024da03f84aa0fcaa87ed4aa1494acb267aa07dc#diff-f35220551ff3accf9393a41a8065a485R140).

Should both be revised?

ansoncfit: "In the linked instance it is using the grid data in the application, so it needs to “fetch” that data."

Indeed, we don't need to "fetch" here.

Issue #578, opened 2017-11-28
Reroute modification bugs


ansoncfit:
Testing https://github.com/conveyal/analysis-ui/pull/575 on staging, there are a few edge cases that aren't being handled:

-When toStop is selected, extendFromEnd should be false.
-When both fromStop and toStop are selected, allowExtend should be false.  Users can edit the geometry between the stop, but they shouldn't be able to extend beyond.

PR #579, opened 2017-11-28
Reroute Fixes

ansoncfit:
Small fixes for #578 

-If only "Start of reroute/extension" (`fromStop`) is set, `extendFromEnd` should be true.
-If only "End of reroute/extension" (`toStop`) is set, `extendFromEnd` should be false (because we want to extend backwards from the start of the reroute segment).
-If both are selected, `allowExtend` should be false.


ansoncfit:
Added some comments.

PR #599, opened 2017-12-11
Fix transit mode selector

trevorgerhardt:
closes https://github.com/conveyal/analysis-backend/issues/97


trevorgerhardt:
whoa, no idea why all those merges showed up here...

PR #602, opened 2017-12-14
Save Analysis Settings

### LINKED ISSUE #594: After deselecting a mode, the button still looks "pressed"
Say I've got the bus mode selected. The button with the bus icon is dark gray like this:
image

If I click it again to deselect this mode, it turns a lighter shade of gray but still darker than other unselected buttons, and still has a blue border around it like this:
image

Only when I click away to another UI element does it take on the appearance I'd expect with all buttons the same shade of light gray:
image

### LINKED ISSUE #600: Date reset when choosing a bookmark
Marco reports:

"When clicking a bookmark, the date of the analysis switches automatically to today. This is not urgent, but was annoying since I had to set the date about 50 times during the meeting this morning."

trevorgerhardt:
Analysis settings are now stored in each project. Each time a travel time surface is requested the settings used for that request are saved to be the defaults for that project. That way the next time a user navigates to the Analysis page, their last used settings for that project will be auto-populated.

Additionally this PR:
- Turns off auto-fetching of results when navigating to the page 
- Makes the mode icon selection more obvious and closes #594.
- Fixes data loading of bookmarks and closes #600


configurations/default/store.yml
```diff
@@ -1,22 +0,0 @@
-mapState:
-  components: []
-  center:
```
ansoncfit: I'm not sure all the places where mapState.center is used.  Will removing it here have unintended side effects?

For example, if I click on Set up a new region, then go back to the main list of regions, then click on Test Region - Boston, then click on Project 1, the scenario editing interface stays centered on Washington DC.  Not a big deal, as this isn't a common sequence, and refreshing the scenario page centers it on Boston.

trevorgerhardt: Is that behavior different than what is on production right now?

ansoncfit: Same as production.

Opened https://github.com/conveyal/analysis-ui/issues/604

lib/components/analysis/bookmark-chooser.js
```diff
@@ -59,6 +59,7 @@ export default class BookmarkChooser extends Pure {
         <div className='row'>
           <div className='col-xs-6'>
             <Select
+              clearable={false}
```
ansoncfit: Agreed, the bookmark selector should not have the option to clear.

But moving the origin/changing the settings from a bookmark should clear the field name.  Otherwise, the settings panel may show a bookmark name that doesn't correspond to the map.

lib/containers/stacked-percentile-selector.js
```diff
@@ -47,19 +48,20 @@ function mapStateToProps (state, ownProps) {
...
     maxAccessibility: selectMaxAccessibility(state, ownProps),
     projectName: `${projectName}: ${variantName}`
   }
 }
 
-function mapDispatchToProps (dispatch: Dispatch, ownProps) {
+function matchDispatchToProps (dispatch: Dispatch, ownProps) {
   return {
-    setIsochroneCutoff: cutoff => dispatch(setIsochroneCutoff(cutoff))
+    setIsochroneCutoff: (cutoff) =>
```
ansoncfit: Adjusting the isochrone cutoff seems laggy.  If this isn't attributable just to using react-dev, maybe the cutoff could be saved to the project only when a new travel time surface is fetched.  We may also just want to have it default to 60 minutes, rather than the last setting used.  Arguments for that approach: it's not really a property of the profile request, and if it's at a low number, users might be confused when they don't see an isochrone.

lib/components/analysis/bookmark-chooser.js
```diff
@@ -59,6 +59,7 @@ export default class BookmarkChooser extends Pure {
...
               options={options}
               disabled={disabled} // don't allow users to reselect bookmarks while their bookmark is loading
```
ansoncfit: It would be nice if previously created bookmarks were selectable before the first analysis was loaded.  This may not be a change from the current version, but the current version automatically made a request on load.

lib/components/analysis/stacked-percentile-selector.js
```diff
@@ -36,6 +36,10 @@ export default class StackedPercentileSelector extends PureComponent {
     setIsochroneCutoff: (cutoff: number) => void
   }
 
+  state: {
+    selection: 'project' | 'base' | 'comparison'
```
ansoncfit: "selection" seems somewhat generic.  "scenarioPlotted"?

Also, should the constants above be used?

lib/containers/__tests__/__snapshots__/single-point-analysis.js.snap
```diff
@@ -248,7 +238,7 @@ exports[`Components > Analysis > Single Point should render correctly 1`] = `
                 />
               </Icon>
                
-              Update results
+              Fetch results
```
ansoncfit: This was changed in messages, right?

lib/components/analysis/bookmark-chooser.js
trevorgerhardt: Yea, I was thinking there should be a way for the UI/"Fetch Analysis" button to know that the settings have changed and new data could be requested. If we implement that it would also be easy to tell if settings have changed from the bookmark.

lib/containers/stacked-percentile-selector.js
trevorgerhardt: Hmmm, so on each "fetch" reset it to 60?

lib/components/analysis/stacked-percentile-selector.js
trevorgerhardt: Can't use the constants here since it's just a flow type. Will update the name to `scenarioPlotted`.

lib/components/analysis/bookmark-chooser.js
ansoncfit: A while ago, the isochrones and accessibility plots would turn grayscale if any of the settings were changed.  Is there a reason that feature was removed?

If not, I can open a separate ticket.

lib/containers/stacked-percentile-selector.js
ansoncfit: Not on each fetch, but maybe each time the page is loaded.

PR #605, opened 2017-12-19
Add flow coverage to remaining tests

evansiroky:


.travis.yml
```diff
@@ -1,19 +1,23 @@
-sudo: false
+addons:
+  apt:
+    packages:
+      - libgif-dev
```
trevorgerhardt: We don't actually use `node-canvas` so this all can be removed along with the `node-canvas` dependency.

evansiroky:
I sorted the `.travis.yml` file manually, but I have used this before: https://github.com/akfernun/yaml-sortkeys

PR #614, opened 2017-12-28
Isochrone Percentile Selection

### LINKED ISSUE #607: Detect if Profile Request settings have changed and reflect that in the UI
Through the Bookmarks, grayscaling the isochrone (what happened to that?), and disabling/enabling the "fetch results" button.

### LINKED ISSUE #611: Minimum probability of change should default to 0 when comparing regional results
We now have worker versions available that do only 1 replication.  The "probability" of change when comparing results from these workers will be very low.  The probability of change slider should default to 0 so that users see the expected accessibility differences on the map.  Unless we have a clearer explanation of the bootstrapping method we use, and the ability to set the number of iterations in the UI, it may make sense to disable the slider.

trevorgerhardt:
Went with a simpler implementation on first pass due to recent discussions so this is kind of a "hidden" feature for now. When you change the percentile of travel time reliability on the single point analysis page it uses that value to select the closest of the predefined values (`[5, 25, 50, 75, 95]`) and uses its value when creating the isochrone. This occurs for displaying and downloading.

This change also defaults the minimum improvement probability on the regional page to 0. Closes #611 

And refactors where we store the single point marker lonlat data. A leftover commit that I didn't notice until I made this PR. Which also detects changes to the Profile Request settings and grays out the isochrone if it's out of date. Closes #607


abyrd:
If you've selected a different percentile, does this also change which line on the cumulative accessibility curve is drawn darker? It was always the 50th (median) before, but if a different value has been selected we should change it.

abyrd:
Also wondering, how hard would it be to just replace that percentile slider with a drop-down list box? I guess that prevents the user from doing anything but our predetermined 5 percentiles, but it might make more sense than rounding it off for isochrone display purposes.

trevorgerhardt:
I'm going to merge this PR and turn the comments above into Issues / further discussions. The reason I went with this simpler, "hidden" style is because we had begun to discuss further changes to the system in Slack like running multiple percentiles at once for Regional Analysis and recommending specific percentiles to use and hiding the rest behind a slider.

abyrd:
@trevorgerhardt I agree with @ansoncfit's point 1. it's going to be a problem though in practice though to have this functionality undocumented/hidden. The label says that the slider only applies to a regional analysis. If someone runs a regional analysis at 25%, then exports some isochrones thinking they are always at 50%, they will in fact be getting incorrect isochrones and accessibility figures. This could really mess up someone's workflow.

PR #615, opened 2017-12-28
Download GeoTIFFs

trevorgerhardt:
Add a button that makes a travel time surface request with `Accept: image/tiff` and downloads the results. If a comparison is selected it requests and downloads both.

### DIFF
--- a/lib/actions/analysis/index.js
+++ b/lib/actions/analysis/index.js
@@ -24,6 +24,7 @@ import {saveToServer as saveProject} from '../project'
 import selectComparisonInProgress from '../../selectors/comparison-in-progress'
 import selectCurrentProject from '../../selectors/current-project'
 import selectCurrentProjectId from '../../selectors/current-project-id'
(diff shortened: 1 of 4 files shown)


lib/actions/analysis/index.js
```diff
@@ -158,6 +166,18 @@ export const fetchTravelTimeSurface = () =>
           } else {
             dispatch(fetchError(responses.value))
           }
+        } else if (asGeoTIFF) {
+          downloadGeoTIFF({
+            data: responses[0].value,
+            filename: `analysis-geotiff-${project ? project.name : ''}-${profileRequest.variantIndex}-${profileRequest.maxTripDurationMinutes}-minutes.geotiff`
```
ansoncfit: The downloaded file will have travel times to all destinations, regardless of the maxTripDurationMinutes selected in the UI, so I'm not sure it makes sense to include that in the filename.

trevorgerhardt: Good call, was copying the filename style from the isochrones.

Issue #609, opened 2017-12-21
isochrone rendering is much slower when no opportunity grid selected


abyrd:
On a Portland transport network, when I have no opportunity grid selected, the isochrone rendering is slow (when I drag the travel time cutoff selector, rendering feels slow and laggy).

Once I select a grid it's totally smooth and fast.

PR #621, opened 2018-01-03
Refactor Selectors

### LINKED ISSUE #618: Group selectors into one file
It's getting messy importing all of the selectors. Group them into one file and re-export. The end result will change importing them from:

```js
import selectProject from '../selectors/project'
import selectCurrentRegion from '../selectors/current-current-region'

function mapStateToProps (state, props) {
  selectProject(state, props)
  selectCurrentRegion(state, props)
}
```

to

```js
import * as select from '../selectors'

function mapStateToProps (state, props) {
  select.project(state, props)
  select.currentRegion(state, props)
}
```

trevorgerhardt:
Cut down on the amount of selector imports. Closes #618


lib/modules/opportunity-datasets/components/dotmap.js
```diff
@@ -3,19 +3,16 @@ import {connect} from 'react-redux'
...
 
-import {
-  selectActiveOpportunityDatasetGrid,
-  selectActiveOpportunityDatasetColorizer
-} from '../selectors'
+import * as select from '../selectors'
 
 function mapStateToProps (state, ownProps) {
-  const grid = selectActiveOpportunityDatasetGrid(state, ownProps)
+  const grid = select.activeOpportunityDatasetGrid(state, ownProps)
```
ansoncfit: Could a quick fix be added here for #609?

lib/selectors/index.js
```diff
@@ -0,0 +1,62 @@
...
+export {default as currentProject} from './current-project'
+export {default as currentRegionId} from './current-region-id'
+export {default as currentRegion} from './current-region'
+export {default as destinationTravelTimeDistribution} from './destination-travel-time-distribution'
+export {default as extendFromEnd} from './extend-from-end'
+export {default as feedScopedStops} from './feed-scoped-stops'
+export {default as feedStopsById} from './feed-stops-by-id'
+export {default as feedStops} from './feed-stops'
+export {default as feedsWithBundleNames} from './feeds-with-bundle-names'
+export {default as fullyQualifiedRouteStops} from './fully-qualified-route-stops'
```
ansoncfit: "fullyQualified" seems a bit vague to me.  Does that mean feed-scoped?

lib/modules/opportunity-datasets/components/dotmap.js
trevorgerhardt: Possibly! I haven't looked into that problem yet so I'm not sure if it's in the scope of this PR.

lib/selectors/index.js
trevorgerhardt: Yes, this one is similar to `feedScopedStops` except it pertains to only the selected modification's route stops. I'll change it so they are the same.

Issue #606, opened 2017-12-19
UI closes itself if user has expired token


evansiroky:
Because of [this line of code](https://github.com/conveyal/analysis-ui/blob/dev/lib/components/application.js#L276), the UI will close the tab in chrome if a user goes to the homepage.  If the token is expired, the first call to the server will return a 403 and cause the error dialog to render.  The last thing to go back to if one opens a new tab is closing the tab in chrome.

Issue #569, opened 2017-11-15
Sort Regions, Projects, Scenarios and Regional Analyses


trevorgerhardt:
Some are sorted by name, some by creation date, some aren't sorted at all. Let's pick one and use that.

Every model will have an `updatedAt` field. We could use that, which would signify the last used item by the client. When they are choosing which one to work on, having that at the top is probably most desireable.

Thoughts @abyrd, @ansoncfit?

ansoncfit:
I would say regions, projects, and regional analyses should be ordered by `createdAt`, with the newest on top.  If we went with `updatedAt`, it could be annoying to have their order changing as different ones were opened/updated.

Scenarios and modifications are currently sorted by `createdAt`, with oldest on top, which I like, especially given the current implementation of numbering scenarios.

abyrd:
Yeah, although it would be nice to have more recently used items on top, it also means they can jump around. I'm not sure how much that would really happen in practice, but sorting by `createdAt` seems good enough.

PR #623, opened 2018-01-05
Application Render Refactor

trevorgerhardt:
Each page is by default wrapped by the `Application` component which contains the `Sidebar` and  the `LeafletMap` amongst other things. This PR refactors a few components to allow each "page" component (each smart component that is supposed to be rendered by a specific route) to compose exactly what it needs on the entire page.

This PR enables us to write components in the new manner and refactors two components as examples of the new style. See `components/create-region` for the ideal style of being able to pass the Map exactly what components and data it needs for the current page.


lib/components/error-modal.js
```diff
@@ -0,0 +1,78 @@
...
+            </a>
+          </li>
+        </ul>
+        <br />
+        <p>
+          <Button
+            block
+            onClick={() => {
+              clearError()
+              window.history.back()
```
ansoncfit: The text of the button says "Close this window", so users might be confused if the button actually takes them back.

See #606 

Doesn't necessarily need to be addressed in this refactor, but I thought I'd mention it.

lib/components/region.js
```diff
@@ -14,7 +14,7 @@ type Props = {
 export default ({children, region}: Props) => region
   ? <div>
     <div className='ApplicationDockTitle'>
-      <Icon type='cubes' /> {region.name}
+      <Icon type='map-o' /> {region.name}
```
ansoncfit: 👍

lib/reducers/region.js
```diff
@@ -9,7 +11,7 @@ export const reducers = {
   'set all regions' (state, action) {
     return {
       ...state,
-      regions: [...action.payload]
+      regions: sortBy([...action.payload], 'name')
```
ansoncfit: Related to #569 ?

I think sorting regions by name makes sense.

PR #636, opened 2018-01-30
Switch from Mapzen's Valhalla to GraphHopper for obtaining street polylines

trevorgerhardt:
Removes all reference/dependence on Mapzen products.


configurations/default/messages.yml
```diff
@@ -82,7 +82,7 @@ region:
     STORING_CENSUS: Storing census data…
     ERROR: Error setting up region...
     DONE: Done
-  customOpenStreetMapData: OpenStreetMap data in PBF format (required for regions outside the USA; OSM extracts are available from Mapzen).
+  customOpenStreetMapData: OpenStreetMap data in PBF format (required for regions outside the USA).
```
ansoncfit: Worth mentioning geofabrik or another extract provider?

trevorgerhardt: We do enough handholding at this stage that I would imagine that this extra advice won't actually help anyone that didn't know where to find OSM data to begin with.

Issue #480, opened 2017-07-28
Key modifiers for trip drawing


chourmo:
When drawing or rerouting a trip, I would greatly appreciate keyboard modifier (shift, alt...) to

- create a stop or a control point
- switch follow route mode

landonreed:
Here are some references in datatools-ui that might be helpful:

- [Help modal documenting keyboard shortcuts](https://github.com/catalogueglobal/datatools-ui/blob/trip-validation-improvements/lib/editor/components/timetable/TimetableHelpModal.js)
- [Gigantic key press handling function](https://github.com/catalogueglobal/datatools-ui/blob/trip-validation-improvements/lib/editor/components/timetable/TimetableGrid.js#L93)

PR #653, opened 2018-03-05
Refactor Edit Geometry

### LINKED ISSUE #574: Disable map panning when dragging route segment to create a control point
I think this is a quick change, maybe in scenario-map/transit-editor/segments.js

### LINKED ISSUE #647: Revised icons for editing alignments
In editing mode, stops show up as subway icons, and control points show up as blue dots.  When editing mode is turned off, the control points disappear and the stops show up as blue dots.  We should revise this inconsistency.

It also seems incongruous to have subway icons representing bus stops.

Perhaps the control points could be smaller blue dots, new stops could be slightly larger circles, and stops that already exist in the network could be the larger circles filled with white.

### LINKED ISSUE #648: Bearing arrows are only shown in one direction for bidirectional trips
image

The bearing arrows we use are a bit messy, and they confusingly only show up in one direction even when modifications are labeled as bidirectional.  

I think I prefer the style of arrows in Data Tools:

image

trevorgerhardt:
Reduce the complexity of the "Transit Editor" code, add instructions when editing, and address a few issues including:

- closes #574 Can no longer drag a route. You can only drag control points and stops.
- closes #647 removes all usage of train icons and replaces them with circular ones.
- closes #648 now only shows arrows for non-bidirectional trips.

screen shot 2018-03-06 at 00 04 24


ansoncfit:
These are very welcome changes!

I'm going to add a few enhancements here, which could be spun out into separate issues.

ansoncfit:
A.
It's still a bit frustrating to add a control point to part of a route that's already created.  Most of the time I try to do this, I end up with my cursor not quite on the line and inadvertently extending a from the end of the route.  This could be mitigated by:

1. Increasing the weight of the alignment, especially in editing mode. (I also think this looks better; the current weight feels too thin compared to the stops).
2. Providing feedback (e.g. using the highlight segment styling) when a user mouses over a segment.
3. Changing "Extend from end" to be trinary: "Extend from: start, end, neither".  Users may need to edit a route geometry but have no intention of extending it.  Being able to click around without spuriously extending would be nice.

ansoncfit:
B. Directional markers could be cleaned up.

I think having them every 125px is too dense https://github.com/conveyal/analysis-ui/pull/653/files#diff-38c540cb5c8a57dd64129696912acdacR46.  Also, could we add some logic so they only show if they're at least a certain distance beyond the start of a segment?  The middle directional marker here is distracting:

image

ansoncfit:
C.
Creating control points could be streamlined.  

Previously, control points could be added just by dragging the route (though this also panned the map [#574]).  Four clicks are now required: clicking on the route, clicking on the newly created stop, making it a control point, and dragging it to the intended location.  I think the previous approach was preferable (especially if the map panning can be disabled).

To add control points while extending also requires extra clicks.  Would it be easy to make ctrl+click add a stop (#480)?

ansoncfit:
D

New segments should be below preceding stops, instead of

image

lib/components/modification/edit-alignment.js
```diff
@@ -150,53 +157,71 @@ export default class EditAlignment extends PureComponent<void, Props, State> {
...
 
         {isEditing &&
           <Checkbox
             defaultChecked={followRoad}
-            label={messages.transitEditor.followRoad}
+            label={message('transitEditor.followRoad')}
             onChange={this._onFollowRoadChange}
           />}
+
+        {isEditing &&
```
ansoncfit: Could we make this collapsible like the timetables below?

ansoncfit:
E
I'm noticing an intermittent bug when dragging a stop.  After clicking an existing stop, when trying to move it, the cursor somehow grabs the leaflet basemap and starts dragging that.  

image

The stop then seems disconnected from the segment:
image

This seems to happen when Follow streets is on.  Initially I thought it was related to the modification auto-saving, but I can't consistently trigger it.

lib/constants/index.js
```diff
@@ -5,6 +5,21 @@
...
 
+/**
+ * Line Editing and display
+ */
+export const GTFS_STOP_COLOR = '#c92336'
+export const AUTO_CREATED_STOP_COLOR = '#888'
+export const CONTROL_POINT_COLOR = '#2389c9'
+export const HIGHLIGHT_SEGMENT_COLOR = '#c96323'
+export const NEW_LINE_COLOR = '#2389c9'
+export const NEW_SNAPPED_STOP_COLOR = '#c92336'
```
ansoncfit: I'm hesitant about using red, since this is also used to represent removed stops

configurations/default/messages.yml
```diff
@@ -129,6 +129,13 @@ transitEditor:
   extendFromEnd: Extend from end
   fromStopLabel: Start of reroute/extension (existing stop on pattern)
   toStopLabel: End of reroute/extension (existing stop on pattern)
+  instructions: Click on the map to create a stop at that location. If "extend from end" is selected, the segment created will extend from the last stop created, otherwise, it will be extended from the first stop. Click on a segment to create a stop at that location. Click on a stop to delete it or change it into a control point. Control points can be used to help position the routes. Click an auto-created stop to turn it into a permanent stop.
```
ansoncfit: I'd make this a bit more concise, something like:
"Click the map to add a new stop in a fixed location. To extend backward from the route's first stop, unselect "Extend from end". Click an existing segment to add a stop to it. Click on a stop to delete it or change it into a control point, which helps define the alignment. Click an auto-created stop to fix its location."

configurations/default/messages.yml
```diff
@@ -129,6 +129,13 @@ transitEditor:
   extendFromEnd: Extend from end
```
ansoncfit: For consistency with the instructions below, we could change "Create stops automatically" to "Auto-create stops"

lib/components/map/draggable-circle-marker.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+import Leaflet from 'leaflet'
+import {CircleMarker} from 'react-leaflet'
+
+export default class DraggableCircleMarker extends CircleMarker {
```
ansoncfit: The old Stop Marker worked well with a touchscreen.  These don't respond to touchscreen or stylus dragging.  If that's an easy fix, using the touchscreen can look cool in demos.

configurations/default/messages.yml
```diff
@@ -129,6 +129,13 @@ transitEditor:
...
   fromStopLabel: Start of reroute/extension (existing stop on pattern)
   toStopLabel: End of reroute/extension (existing stop on pattern)
+  instructions: Click on the map to create a stop at that location. If "extend from end" is selected, the segment created will extend from the last stop created, otherwise, it will be extended from the first stop. Click on a segment to create a stop at that location. Click on a stop to delete it or change it into a control point. Control points can be used to help position the routes. Click an auto-created stop to turn it into a permanent stop.
+  instructionsTitle: Editing Instructions
+  existingStops: Existing stops from GTFS Feed.
+  newStopDescription: A new stop.
```
ansoncfit: A/an at the start of these legend items, and periods at the end, probably aren't necessary

lib/constants/index.js
trevorgerhardt: Kind of agree, mostly was going for high contrast with the blue. Here's a color palette generated from Conveyal blue (#2389c9): https://palx.jxnblk.com/2389c9

Alternate suggestions?

lib/components/map/draggable-circle-marker.js
trevorgerhardt: Erg. Yea, regular markers are more appropriate and a quick try of getting CircleMarkers to work with touch events did not work. But there's an issue I haven't resolved with DivIcon offsets when they're small...

J/k I just resolved it by not using SVG icons and only using CSS. Will push a fix soon.

lib/components/modification/edit-alignment.js
trevorgerhardt: Yes, but is it necessary? I was considering hiding the timetables/other non-editing options while in editing mode also.

ansoncfit: Not necessary, but I thought it would be worth it to get the instructions out of the way of the timetable entries, and to condition users to expanding/collapsing the timetable entries.

lib/constants/index.js
ansoncfit: Either the same as GTFS_STOP_COLOR (which have a different size), or one of the grays from the palette you suggested (#364046)?

trevorgerhardt: Hmmm, I was trying to place more emphasis on the fact that the stops are snapped...gray is less emphasis. `GTFS_STOP_COLOR` is currently #c92336 (red 5 from that palette)

ansoncfit: Oops,  GTFS_STOP_COLOR is indeed red.  They are small enough that they looked gray to me.

Comments in /lib/constants/colors.js say "added things are blue", "removed things are red", and "neutral things are gray."  In my mind, existing stops from the GTFS feed are neutral -- they cannot be manipulated.  And I think if the gray were dark enough, these would stand out as snapped/transfer stops.

For writing docs, I like the simple approach outlined by the comments in colors.js and would like to stay consistent with that.  I guess this also raises the question of why these constants are here rather than in the colors file.

trevorgerhardt:
@ansoncfit with regards to A I'm going to:

- Increase the line weight based on the zoom level
- Have a stop appear under the mouse when you are hovering over an empty part of a segment

I played around with having a line to your pointer show and update on mouse move. Seemed too obtrusive for now but with the option to turn it off with the trinary control it might be fine. I would say let's combine those changes in a subsequent PR.

trevorgerhardt:
B. Directional markers could be done better, but preferably in a subsequent PR.

trevorgerhardt:
C. I like the idea of still requiring a click on the segment but using a keyboard option to create a control point instead. Then it's click and drag.

D & E should be fixed on the next push 👌

lib/constants/index.js
trevorgerhardt: Ok, for consistency's sake that makes sense. I'm adding an orange "highlight" color also.

trevorgerhardt:
@ansoncfit addressed all of the new problems, changed the colors as discussed, updated the messages to your recommendations.

ansoncfit:
@trevorgerhardt A1 and A2 look great.  Thanks for trying A3, let's address later (#657 )

Agreed, let's also revisit B (#658) and C later.

Looks like we're ready to merge once snapshots are updated.

PR #661, opened 2018-03-26
Update Auth0

### LINKED ISSUE #642: Cannot log out from landing screen
On the initial analysis page that lists projects, there's no way to change to another user. You have to select a project then log out which is not an obvious sequence of actions.

trevorgerhardt:
This is a high priority update as Auth0 changed the way they handle requests (needs to be deployed to production by April 1st).

We now redirect to `/login` before making any requests to Auth0. So each page load will redirect there, check the session, and redirect back to the page you are requesting. For the most part that will only need to happen once per session.

Also closes #642


evansiroky:
I looked at the code and it seems reasonable.

PR #663, opened 2018-03-27
Update parsing order of time grids

trevorgerhardt:
BREAKING CHANGE: Incompatible with old grid data format


lib/actions/analysis/parse-times-data.js
```diff
@@ -0,0 +1,99 @@
...
+
+type TimesData = {
+  version: number,
+  zoom: number,
+  west: number,
+  north: number,
+  width: number,
+  height: number,
+  nSamples: number,
+  data: number[]
```
abyrd: This is stating that the data and dimensions are JS numbers which is not entirely accurate - they're constrained to be integers and everything is designed such that they could even be 8-bit integers if we need to make that optimization to handle large regions. See comments below about how the data are stored.

lib/actions/analysis/parse-times-data.js
```diff
@@ -0,0 +1,99 @@
...
+    let previous = 0
+    for (let j = 0; j < sampleSize; j++) {
+      const stored = data[offset++]
+      const current = stored + previous
+      times.push(current)
+      previous = current
+    }
+  }
+
+  const decoder = new TextDecoder('utf-8') // utf-8 is Jackson default
```
abyrd: It's not just Jackson's default, it's a near-requirement of the format itself. See https://tools.ietf.org/html/rfc8259 §8.1: "JSON text exchanged between systems that are not part of a closed ecosystem MUST be encoded using UTF-8 [RFC3629]."

lib/actions/analysis/parse-times-data.js
```diff
@@ -0,0 +1,99 @@
...
+    height,
+    nSamples,
+    data: times,
+    errors: [],
+    warnings: metadata.projectApplicationWarnings || [],
+    get (x: number, y: number) {
+      const samples = []
+      const basePosition = (y * width + x)
+      for (let i = 0; i < nSamples; i++) {
+        samples.push(times[(i * sampleSize) + basePosition])
```
abyrd: Ah, this might be the explanation for the ordering of the original format: every time we call this new form of this get() function we're reading from N disparate places in memory. This could be significantly slower than the original ordering, depending on how heavily this get() function is called. I need to check where this is being called - why would we need all N samples at once to make a single isochrone or a single cumulative accessibility curve? It seems like these days we're treating each plane of the travel time results independently. This function could be an artifact of the old element ordering.

abyrd:
@trevorgerhardt should I give this another look? Is it intended to be merged soon? I see that the Travis build is failing for some reason.

trevorgerhardt:
@abyrd I would rather not merge this until we're ready to deploy the latest r5/backend since it contains breaking changes. I also need to update the description with the entire list of changes that this contains since the scope of this PR has expanded.

ansoncfit:
Modal dialogs are strange on this branch:

image

ansoncfit:
The above does seem like it was just a transient error, fixed after restarting the front-end.

I'm in favor of merging.

PR #676, opened 2018-04-20
Add validation to number input

### LINKED ISSUE #652: Don't allow users to enter 0 for walk/bike speed
We should also check this on the back-end (see https://github.com/conveyal/r5/issues/410)

evansiroky:
Fixes #652


lib/components/input.js
```diff
@@ -136,18 +137,71 @@ const preventDefaultIfFocused = e => {
...
+    if (!error) {
+      onChange(e)
+    }
+  }
+
+  _renderHelpBlock () {
+    const {max, min} = this.props
+    let helpText = ''
+    if (isNumber(min)) {
+      helpText += min
```
trevorgerhardt: Would it be more concise if the help text was generated when the error was found? Are we showing help text when there is no error?

lib/components/input.js
```diff
@@ -136,18 +136,63 @@ const preventDefaultIfFocused = e => {
...
+    }
+    const maxVal = parseFloat(max)
+    if (!isNaN(maxVal) && value > maxVal) {
+      error = true
+    }
+    this.setState({error})
+    if (!error) {
+      onChange(e)
+    }
+  }
```
trevorgerhardt: Let's rewrite this to:
```js
const value = parseFloat(e.target.value)
const max = parseFloat(this.props.max) 
const min = parseFloat(this.props.min || 0) // can assume this will always be a number 
if (isNaN(value) || value < min || (!isNaN(max) && value > max)) {
  return this.setState({error: true})
}
this.setState({error: false})
this.props.onChange(e)
```

lib/components/input.js
```diff
@@ -136,18 +136,63 @@ const preventDefaultIfFocused = e => {
...
+
   render () {
     const id = propsToId(this.props)
+    const {className, onChange, ...propsLessOnChange} = this.props
+    const {max, min = 0} = this.props
+    const {error} = this.state
+    const groupProps = {...this.props}
```
trevorgerhardt: Do we want `className` just going to `groupProps`? Also, it appears to get `onChange` here.

evansiroky: The Group component doesn't have any change event so that shouldn't matter.  If we need to we could create new props for the group className and input className.

Issue #544, opened 2017-11-03
User should be warned if required modification values are not entered


ansoncfit:
We could figure out some simple sense checking for each modification type (e.g. required fields). If conditions are not met:

- An exclamation point could show up next to the back button
- Clicking on the back button could trigger an alert warning the user and giving the option to leave the modification detail anyway, or cancel.

Related: visual styling should emphasize which fields are required.

PR #677, opened 2018-04-21
Add trip info helpers

evansiroky:
Warn user of required input in add trip modifications.  See #544


lib/components/modification/edit-alignment.js
```diff
@@ -135,6 +137,46 @@ export default class EditAlignment extends PureComponent {
...
+    modification.segments.forEach(segment => {
+      if ((lastSegment && lastSegment.toStopId !== segment.fromStopId) ||
+        (!lastSegment && segment.fromStopId)) {
+        // make sure we don't double count stops
+        numStops++
+      }
+      if (segment.toStopId) {
+        numStops++
+      }
+      distance += lineDistance(segment.geometry, 'kilometers')
```
trevorgerhardt: The modification component already has `segmentDistances` calculated. You can pass that into `EditAlignment` and add up from there. Could also create a `stop-count` selector to fit that pattern. 

https://github.com/conveyal/analysis-ui/blob/dev/lib/components/modification/add-trip-pattern.js#L24
https://github.com/conveyal/analysis-ui/blob/dev/lib/selectors/segment-distances.js

lib/components/modification/edit-alignment.js
```diff
@@ -135,6 +137,46 @@ export default class EditAlignment extends PureComponent {
...
+      lastSegment = segment
+    })
+
+    if (distance === 0) {
+      return (
+        <div className='block'>
```
trevorgerhardt: Are these blocks needed? The screenshots posted in Slack are overly padded.

lib/components/modification/edit-alignment.js
```diff
@@ -135,6 +137,46 @@ export default class EditAlignment extends PureComponent {
...
+          <div className='alert alert-danger' role='alert'>
+            <span className='glyphicon glyphicon-exclamation-sign' aria-hidden='true' />
```
trevorgerhardt: Any reason you aren't using `<Icon type='exclamation-sign' />` here?

lib/components/modification/edit-alignment.js
```diff
@@ -135,6 +137,46 @@ export default class EditAlignment extends PureComponent {
...
+            <span className='sr-only'>Error:</span>
```
trevorgerhardt: Is this an `Error`? I would label it differently.

lib/components/modification/timetables.js
```diff
@@ -71,6 +71,24 @@ export default class Timetables extends PureComponent<void, Props, void> {
             <Icon type='plus' /> Add timetable
           </Button>
         </p>
+        {timetables.length === 0
+          ? (
+            <div className='block'>
+              <div className='alert alert-danger' role='alert'>
+                <span className='glyphicon glyphicon-exclamation-sign' aria-hidden='true' />
+                <span className='sr-only'>Error:</span>
```
trevorgerhardt: As commented above, is this really an error?

lib/components/modification/timetables.js
```diff
@@ -71,6 +71,24 @@ export default class Timetables extends PureComponent<void, Props, void> {
             <Icon type='plus' /> Add timetable
           </Button>
         </p>
+        {timetables.length === 0
+          ? (
+            <div className='block'>
+              <div className='alert alert-danger' role='alert'>
+                <span className='glyphicon glyphicon-exclamation-sign' aria-hidden='true' />
```
trevorgerhardt: Same as above, use `Icon`

lib/components/modification/timetables.js
```diff
@@ -71,6 +71,24 @@ export default class Timetables extends PureComponent<void, Props, void> {
...
+                <span className='sr-only'>Error:</span>
+                At least 1 timetable must be added
+              </div>
+            </div>
+          )
+          : (
+            <div className='block'>
+              <div className='alert alert-info' role='alert'>
+                {`${timetables.length} timetable(s)`}
```
trevorgerhardt: This seems like it should be part of a section heading instead of an info box: `Timetables (${timetables.length})`

lib/components/modification/edit-alignment.js
evansiroky: Ah, didn't realize that there were selectors for these things.

lib/components/modification/edit-alignment.js
evansiroky: I just used the standard bootstrap v3 html for wells and icons.  @ansoncfit and you and I should have a quick slack call to discuss the styling.  This comment applies to the rest of the comments you made in this PR.

lib/components/modification/edit-alignment.js
```diff
@@ -135,6 +138,43 @@ export default class EditAlignment extends PureComponent {
     }
   }
 
+  _getTripPatternHelperText (): { error: boolean, textHtml: any} {
```
trevorgerhardt: Oi. I'm not a fan of this pattern.

lib/components/modification/edit-alignment.js
```diff
@@ -146,9 +186,12 @@ export default class EditAlignment extends PureComponent {
     const hasAnyPhasing =
       modification.timetables &&
       modification.timetables.findIndex(t => t.phaseAtStop != null) !== -1
+    const {error, textHtml} = this._getTripPatternHelperText()
 
     return (
       <div>
+        <h5>Route Geometry</h5>
+        {!error && textHtml}
```
trevorgerhardt: I think the logic in the `_getTripPatternHelperText` is pretty muddled. With the distance, we could just replace this part with
```js
{distance > 0 && type === ADD_TRIP_PATTERN && <p>{`${numberOfStops} stops over ${Math.round(distance * 100) / 100} km`}</p>}
```

lib/components/modification/edit-alignment.js
```diff
@@ -160,6 +203,8 @@ export default class EditAlignment extends PureComponent {
             <Icon type='stop-circle' /> {message('transitEditor.stopEdit')}
           </Button>}
 
+        {error && textHtml}
```
trevorgerhardt: And this one with
```js
{distance === 0 && type === ADD_TRIP_PATTERN && 
  <div className='alert alert-danger' role='alert'>
    <Icon type='exclamation-circle' /> A route geometry must have at least 2 stops
  </div>}
```

lib/components/modification/timetables.js
```diff
@@ -66,11 +66,20 @@ export default class Timetables extends PureComponent<void, Props, void> {
...
     return (
       <div>
+        <h5>Timetables ({timetables.length})</h5>
         <p>
           <Button block onClick={this._create} style='success'>
             <Icon type='plus' /> Add timetable
           </Button>
         </p>
+        {timetables.length === 0 &&
+          <div className='block'>
```
trevorgerhardt: This `<div className='block'>` shouldn't be necessary as this section is already padded.

lib/components/modification/edit-alignment.js
evansiroky: That is a lot cleaner and concise

lib/components/modification/timetables.js
evansiroky: ah, yes.  I went ahead and applied the same padding within a `p` element to the edit-alignment component.  Looks better now.

trevorgerhardt:
LGTM, failing lint is from the branch we're merging into.

Issue #679, opened 2018-04-24
Allow renaming of regional analyses


ansoncfit:

ansoncfit:
Resolved

Issue #590, opened 2017-12-05
Add a `confirm` when deleting a completed regional analysis


trevorgerhardt:
Even though it's soft delete, it happens way too easily

PR #685, opened 2018-05-08
Regional Analysis Enhancements

trevorgerhardt:
- Addresses #679 
- Allows navigating to an incomplete Regional Analysis page without throwing an error
- Shows completed Regional Analyses in a Select box


lib/components/analysis/regional-results.js
```diff
@@ -375,131 +352,108 @@ export default class RegionalResults extends Component<void, Props, State> {
...
-          {analysis ? analysis.name : 'Loading...'}
-          <IconLink
-            className='pull-right'
-            onClick={this._deleteAnalysis}
-            title={messages.analysis.deleteRegionalAnalysis}
-            type='trash'
+        {this.renderComparisonSelector()}
+
+        {comparisonAnalysis &&
+          <MinimumImprovementProbabilitySelector
```
ansoncfit: Recent worker versions have 0 bootstrap replications, so I think we should consider hiding this selector because it doesn't work as intended.

lib/components/analysis/regional-results.js
```diff
@@ -375,131 +352,108 @@ export default class RegionalResults extends Component<void, Props, State> {
...
+          </Button>
+          {comparisonAnalysis &&
+            <Button block onClick={this.downloadComparisonGIS} style='info'>
+              {sprintf(
+                messages.analysis.downloadRegional,
+                comparisonAnalysis.name
+              )}
+            </Button>}
+          {comparisonAnalysis &&
+            <Button block onClick={this.downloadProbabilityGIS} style='info'>
```
ansoncfit: As with previous comment, we should consider disabling this unless bootstrap replications are present.

lib/components/analysis/regional.js
```diff
@@ -0,0 +1,88 @@
...
+
+  componentDidMount () {
+    const p = this.props
+    p.loadRegionalAnalyses(p.regionId)
+    p.setActiveRegionalAnalysis({_id: p.analysisId})
+  }
+
+  _deleteAnalysis = () => {
+    const p = this.props
+    p.deleteAnalysis(p.analysisId, p.projectId)
```
ansoncfit: Would now be a good time to address https://github.com/conveyal/analysis-ui/issues/590?

lib/components/analysis/regional-analysis-selector.js
```diff
@@ -13,107 +13,59 @@ type Props = {
...
-            <Analysis
-              analysis={a}
-              key={`analysis-${a._id}`}
-              onDelete={this._deleteRegionalAnalysis(a._id)}
-              onSelect={this._selectRegionalAnalysis(a._id)}
+      <div>
+        <Group>
+          <Select
+            onChange={this._goToRegionalAnalysis}
+            options={allAnalyses.map(a =>
```
ansoncfit: Is this sorted alphabetically by name?

I would be in favor of `allAnalyses` being limited to completed analyses for this project.  It might also be helpful to have a separate dropdown with completed analyses for other projects in the region, but I'm not sure of that.

trevorgerhardt: No, this is sorted by "createdAt". You can now see go to an RA that is not complete yet. It could be useful to see more detailed information about it that way.

ansoncfit: Agree it could be useful to see more detailed information.  At first glance, I thought running jobs were all being shown below this dropdown, in which case it would be duplicative to have them in the dropdown as well.

PR #691, opened 2018-05-16
Correctly pass analysis errors back to the UI

trevorgerhardt:
Fixes a bug that was caused during the renaming process.


ansoncfit:
For background, we used to have separate ways of reporting scenario errors (e.g. due to malformed modifications) and warnings (e.g. routes without any trips specified in baseline GTFS).  In the old system with a separate broker, errors would short-circuit routing and be reported in the UI.  

In the new system (see https://github.com/conveyal/r5/pull/425), errors are reported back in the same way as warnings along with a time grid filled with `UNREACHABLE`.  I'm not sure `errors[]` would ever be given the current backend/worker setup.  We should check to see if this short-term fix (which is admittedly not ideal) works on staging.

PR #703, opened 2018-06-15
Copy timetable fixes

### LINKED ISSUE #701: Timetables to copy don't appear unless fields reselected
It's not obvious that one needs to change the modification value to get timetables to show up:
image

And if there's only one modification, it's not possible to reselect it, so the user has to reset the project/region.

Perhaps the fields should default to "Select a project" "Select a modification" etc.  This could also address the issue of the modal just sticking on "Loading" if there are no timetables to copy.

### LINKED ISSUE #702: Copied timetables should include speed

evansiroky:
- Properly render selectors so there is no confusion of which timetable is possible to select.  (Fixes #701)
- Show warnings when segments differ between modifications when copying timetables.  (Fixes #702)
- Create new timetable with unique name.  Ex:  "Copy of timetable x".

### DIFF
--- a/lib/components/modification/copy-timetable.js
+++ b/lib/components/modification/copy-timetable.js
@@ -2,7 +2,6 @@
 
 import Icon from '@conveyal/woonerf/components/icon'
 import get from 'lodash/get'
-import ordinalSuffix from 'ordinal-suffix'
 import React, {Component} from 'react'
 import uuid from 'uuid'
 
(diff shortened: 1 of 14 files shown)


lib/components/modification/copy-timetable.js
```diff
@@ -237,64 +278,75 @@ export default class CopyTimetable extends Component {
...
+            </option>
+          ))}
+        </Select>
+        {modification &&
+          currentModification.segments.length !== modification.segments.length &&
+          <div className='alert alert-info' role='alert'>
+            <Icon type='exclamation-circle' />
+            {(currentModification.segments.length === 0 ||
+              modification.segments.length === 0) &&
+              sprintf(
```
trevorgerhardt: Instead of `sprintf` and `utils/messages` use [woonerf/message](https://github.com/conveyal/woonerf#message). We haven't implemented it everywhere, but all new code should use it.

```js
import message from 'woonerf/message'
...
message('modification.copyTimeTable.noSegments', {speed: DEFAULT_SEGMENT_SPEED})
```

Issue #649, opened 2018-02-28
UI shows no error detail


abyrd:
I get an error popup that says "Unfortunately it appears an error has occured... Error on the server" with no details at all. If I look in the developer tools I see that there is in fact an XML payload:

```XML
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Request has expired</Message><Expires>2018-02-28T08:10:07Z</Expires><ServerTime>2018-02-28T08:10:12Z</ServerTime><RequestId>3231326307A2B4CA</RequestId><HostId>kmiEXmyD7C5d2a2i2JEeYqGb0+/cm0qhng6piUF6UXZklB30dsYCuyqUYcVErjvpzuzexZ3fUUM=</HostId></Error>
```

But this is not displayed in the popup in the UI.

abyrd:
I'm seeing this in error reports from users as well - they send us a screen capture of a popup that says "error on the server" without the response body, which presumably contains some more detail. My guess would be that the UI is trying to parse the body as JSON but some errors are plain text or XML depending on where the failure happens in the server. If that's where the problem lies, we should make an effort to display the raw response body whenever JSON parsing fails.

Issue #545, opened 2017-11-03
Support email should not be hard-coded


ansoncfit:
We may want to specify the support email in the same place we specify API keys, instead of https://github.com/conveyal/analysis-ui/blob/c9a53e9f74386de40450d5d9e4d3b0f852b14b0a/lib/components/application.js#L245

This would make sure that our support email address only shows up for our supported deployments.

Issue #491, opened 2017-08-29
Upload shapefile button kind of hard to find


abyrd:
It took me a while to realize it's the globe icon at the top of the scenario, which situates it as a kind of global operation rather than something that adds individual modifications to the scenario. It's not absurd, but just feels a bit counterintuitive.

It also has an obscure icon. Personally I hesitate to use any non-obvious icons. Considering how rarely many people will use this, and how unlikely they are to make a permanent association between a picture of a globe and uploading a shapefile of transit lines, I think it makes more sense to have a button that says "upload route alignment shapefile".

ansoncfit:
We could use fa-upload to upload a shapefile of a scenario, and fa-download to go to the dialog for downloading a variant .json (and maybe eventually a variant .shp, as suggested in #255)

PR #707, opened 2018-06-30
UX updates

ansoncfit:
This PR:
- Cleans up error windows, with full stack traces from the server in a collapsible div (addresses #649): 
image
- Declutters the project title bar.  Most of the buttons there were rarely used.  Deleting a project is now an option on the project settings page (which should help avoid inadvertent clicks).  Importing route alignments from a shapefile can now be accessed after clicking on the import modifications button (addresses #491)
- Reduces the number of buttons shown above the projects list from four to one.  Region settings and Opportunity Dataset panels can be accessed from the sidebar; having duplicate buttons in the project list could be confusing.  Now either Upload GTFS Bundle or Create New Project is shown, depending on whether a bundle exists.
- Refactors messages to use woonerf (including progress toward #545).
- Fixes a bug that prevented phasing (#708)
- Changes redirection when deleting a modification.  But we may want to revert this commit given the larger changes to routes in the `refactor-analysis` branch, which I should have discussed before making this change.  I also noticed that users are not pushed to a modification details page when a new modification is created.


trevorgerhardt:
👍 

Looks like there are minor lint errors that need to be fixed and then the snapshots will need to be updated but overall, this looks good.

PR #712, opened 2018-07-19
Refactor Opportunity Datasets for better management

trevorgerhardt:
- Edit, delete entire source, and allow for easier changes in the future due to decoupling from regions and managing as their own database model.
- Manually download all LODES data for a region

Requires https://github.com/conveyal/analysis-backend/pull/152


ansoncfit:
We just decided to add a button for fetching seamless-census data upon request, rather than automatically at region creation.  We should specify which year of LODES data we're fetching in a minimal number of places for easier updating.  For example, in analysis-backend, we'll want seamless-census-bucket to be lodes-data-2015 for now, then we'll want to update in September once new data are published.  Maybe the frontend button uses a message ("Fetch latest LODES data (2015)")

lib/constants/colors.js
```diff
@@ -56,5 +56,14 @@ export default {
     'rgba(255, 255, 191, 0.42)',
     'rgba(171, 217, 233, 0.42)',
     'rgba(44, 123, 182, 0.42)'
+  ],
+
+  OD_GRADIENT: [
```
ansoncfit: To avoid future origin-destination confusion, could we make this explicitly OPPORTUNITY_DATASET_GRADIENT?

lib/modules/opportunity-datasets/components/choropleth.js
```diff
@@ -0,0 +1,22 @@
+// @flow
```
ansoncfit: Does this do what I think it does? 😃

lib/modules/opportunity-datasets/components/list.js
```diff
@@ -124,58 +96,58 @@ export class ListOpportunityDatasets extends PureComponent {
...
-          title={messages.opportunityDatasets.upload}
-          ><Icon type='plus' /> Upload a new dataset
+          title={message('opportunityDatasets.upload')}
+          ><Icon type='plus' /> {message('opportunityDatasets.upload')}
+        </Button>
+        <Button
+          block
+          onClick={this._downloadLODES}
+          style='primary'
+          title={message('opportunityDatasets.downloadLODES', 'Download LODES for Region')}
```
ansoncfit: I couldn't find this in the messages file, but somehow it's getting set to "Download LODES".

I'd suggest "Import LODES (2015)" or "Fetch LODES (2015)" and different fa-icon; user isn't downloading to their machine.

lib/modules/opportunity-datasets/components/list.js
```diff
@@ -124,58 +96,58 @@ export class ListOpportunityDatasets extends PureComponent {
...
-      <p>Finished uploading {status.totalFeatures} features {distanceInWords(status.completedAt)} ago. <a onClick={() => {
-        status.clear()
-        window.location.reload(true)
-      }} tabIndex={0}>Reload page</a> to see results</p>}
-    {!isErrorOrDone(status.status) && <p>Started {distanceInWords(status.createdAt)} ago and completed {status.completedFeatures}/{status.totalFeatures} features.</p>}
+      <p>
+        {message('opportunityDatasets.finishedUploading', {
+          total: `${status.totalGrids}`,
+          completedAt: distanceInWords(status.completedAt)
+        })}
```
ansoncfit: Minor: last status not cleared on completion:

image

lib/modules/opportunity-datasets/components/choropleth.js
trevorgerhardt: I tested it! It's available, not currently used anywhere.

lib/constants/colors.js
trevorgerhardt: 👍

lib/modules/opportunity-datasets/components/list.js
trevorgerhardt: 🤔 It can be cleared with the "X". Or it gets cleared automatically it's over 7 days old.

lib/modules/opportunity-datasets/components/choropleth.js
ansoncfit: This would have been useful for a recent customer request; looking forward to using  🌈

lib/modules/opportunity-datasets/components/list.js
ansoncfit: I meant the "Extracting census data for region" part.  Not worth spending much time on.

lib/modules/opportunity-datasets/components/choropleth.js
ansoncfit: I meant just on the opportunity dataset page.  Agree a choropleth map wouldn't really work under an isochrone.

lib/modules/opportunity-datasets/components/list.js
trevorgerhardt: Ah.

Issue #696, opened 2018-05-29
Show all selected routes/patterns in modifications


ansoncfit:
For adjust-speed modifications that include multiple routes, only the first route shows on the map and in the printable report.

ansoncfit:
Upon further investigation, it turns out that multiple patterns can be shown on the map, but but only if they have been shown individually first.

To reproduce this:

1. Create an `allowMultipleRoutes` modification (e.g. adjust speed).
1. Select route 1.  It will show on the map.
1. Add to the selection route 2.  It will not show on the map.
1. Remove from the selection route 1.  Route 2 will show on the map.
1. Add route 1 back to the selection.  Both routes will show on the map.
1. But upon refreshing the page or loading the printable report, only one modification will be shown.

trevorgerhardt:
This most likely has to do with how we determine what patterns to fetch using via the GraphQL/GTFS api. I'd look into `actions/get-feeds-routes-and-stops.js`

ansoncfit:
Ah yes, there's an existing TODO:

https://github.com/conveyal/analysis-ui/blob/dev/lib/actions/get-feeds-routes-and-stops.js#L163

Issue #457, opened 2017-07-20
Remove trips for multiple trips


chourmo:
When editing a large network, it would simplify variant sets if it was possible to delete several trips in the same "remove trips"

ansoncfit:
To clarify, multiple patterns can be selected for a given route.

But multiple routes cannot be selected in one remove trip modifications.  It would be great if multiple routes, or even all routes in a given feed, could be selected.

ansoncfit:
This was also requested by PUC users a couple months ago.

PR #713, opened 2018-07-23
Small fixes for modifications with multiple routes

ansoncfit:
- [x] #457 
- [x] #696


trevorgerhardt:
@ansoncfit I've never looked further into this PR because the tests fail and there's another checkbox waiting to be checked. Should I review it? Will this work be continued?

ansoncfit:
I just added more notes to #696.  I could use some guidance on how to approach a fix for that, @trevorgerhardt.

PR #717, opened 2018-07-26
Allow importing Add Trips modifications

### LINKED ISSUE #645: Allow import of add-trip modifications that were created using different bundles
Currently, the ability to import modifications from another project is restricted to projects that use the same bundle.

It would be useful to be able to import add-trip modifications, either via upload or from projects that use different bundles.  I think this would be fairly straightforward, requiring an update to the `projectId` and the rewriting of all `fromStopId` and `toStopId` values to `null`.

This could also be a step toward more fully replacing the old geom2gtfs, which allowed importing both alignments and stop locations.  We currently only allow alignment import, allowing users to set stop spacing values and requiring them to manually drag stops if they want them in specific locations.

trevorgerhardt:
Allows selecting projects that have different bundles and adds a note explaining what will happen. Import occurs on the backend. Closes #645 

Relies on https://github.com/conveyal/analysis-backend/pull/153


configurations/default/messages.yml
```diff
@@ -178,6 +178,7 @@ modification:
   clearBundleConfirm: Clear selected bundle
   importFromShapefile: Route alignments from shapefile
   importFromProject: Modifications from another project
+  importFromProjectInfo: Projects with different bundles will only have "Add Trips" modifications copied and those modifications will have stop references removed.
```
ansoncfit: Suggest "When importing from a project associated with a different bundle, only "Add Trips" modifications will be copied. If a modification uses an existing stop from the baseline GTFS, it will be replaced in the imported modification by a new stop at the same location."

lib/components/import-modifications.js
```diff
@@ -67,11 +67,12 @@ export default class ImportModifications
           </Button>
           <br />
           <h5>{message('modification.importFromProject')}</h5>
+          <p>{message('modification.importFromProjectInfo')}</p>
```
ansoncfit: This message is a bit verbose (and my suggestion makes that worse).  Is there an easy way to hide this unless a project with a different bundle has been selected?

PR #749, opened 2018-09-11
Update mastarm

trevorgerhardt:
Depends on https://github.com/conveyal/mastarm/pull/245

WARNING: This currently turns off flow type checking because it never activates correctly locally for me. Appears to be too many files and I'm not entirely sure how to manually ignore enough. Ignoring all of node_modules is not possible because then we get errors that libraries are missing.

This is a first step in updating all other dependencies (React to v16, which needs to be done). Some other items of significance:

- Drops all direct usage of `sprintf.js` and uses `woonerf/message` instead. This required lot's of small changes to `messages.yml`.
- Ran `mastarm lint lib/ --fix`. With new rules, this will make lot's of small ordering and whitespace changes.
- New versions of testing frameworks caused almost all snapshots to be updated (Ex: `Function` -> `MockFunction`)
- Component's must have `<Props>` now, so all `props: Props` had to be changed.
- The `Event & {target: HTMLInputElement}` style was giving more flow errors (although flow is now currently off) and in my research I found that we should've been using `SyntheticEvent<?>`'s anyway. I'm not 100% sure that I'm doing it correctly, but it's more "appropriate" than before. See Flow docs for more info: https://flow.org/en/docs/react/events/
-


trevorgerhardt:
Going to close this and do each part incrementally so that it's actually testable.

Issue #756, opened 2018-09-17
Percentile slider says "regional only"


ansoncfit:
We could display both rounded value and the exact value.  E.g.:

"Travel time percentile: 25 (single-point), 20 (regional)"

trevorgerhardt:
Added back...but we now say the same thing in slightly different ways. See 8268e4f933a0e5663584f155961310f93fad12a7

Issue #737, opened 2018-08-09
Downloadable JSON with regional analysis settings


ansoncfit:
Requested by a users:

"Ability to export a text log file for regional analyses that can assist with documentation (contains key information such as modes, times, simulated schedules, scenarios etc,)"

This ideallly would include the scenario JSON itself.

ansoncfit:
The currently saved scenario json is in R5 format. 

We could also snapshot the modifications in front-end format to allow printable reports.

abyrd:
Saving the modifications in the non-R5 format (used by both the UI and backend) would also be helpful in resolving https://github.com/conveyal/analysis-backend/issues/214. With the new custom modification concept, the possibility exists of modifications that are understood by some version of R5 workers but cannot be deserialized by the backend.

trevorgerhardt:
Closed in #1156

Issue #619, opened 2018-01-02
Darken selected percentile curve line on cumulative opportunity plot


trevorgerhardt:

ansoncfit:
We could also display the corresponding accessibility values in the progress bars above.

trevorgerhardt:
This is a nice idea, but I would rather not make the cumulative accessibility plot any more complex.

ansoncfit:
My comment above is implemented -- the progress bars do update to reflect the selected percentile.

I do think the cumulative curves themselves would be much easier to interpret if they somehow reflected the selected percentile. Is the concern about complexity from a user perspective, or from a code/maintenance perspective?

ansoncfit:
Similar comment for the revised box plot introduced in #1183: indicating the value corresponding to the selected percentile (e.g. with a double-width stroke) would make it easier to understand what the displayed numbers to the left of the boxplot represent.

Issue #538, opened 2017-11-02
Show more information for completed regional analyses


ansoncfit:
We should display more information for completed regional analyses:
 - R5 version used
 - Egress modes
 - Scenario name used for analysis, potentially linked to downloadable JSON.  We need to figure out how to make sure that if a scenario is updated after a regional analysis has been run, there is a way to figure out what the original scenario was when looking at regional results.

ansoncfit:
We're close to closing this issue, but we need to figure out the last bullet point.

The details table should have a row for "Scenario", and the value should be the name of the scenario.  I propose checking to see if the scenario used for the analysis is exactly the same as the scenario as it is currently specified (using hashes for scenarios?).  If they are the same, the scenario name could link to the modifications page.  If not, I would propose removing the link from the name, and adding a warning icon, with on hover text: "The scenario used to run this analysis has been altered; you can see details in the downloadable analysis log" (see #737)

ansoncfit:
It could be misleading to set the scenario name based on the project Id and variant Index when regional results are displayed, since the scenario might have been renamed or altered since the regional analysis was run.  As an alternative, we could just suggest that the name of a scenario at the time it is run be included in the regional analysis name.

abyrd:
Maybe we should just save the entire request in the regional analysis database entry.

trevorgerhardt:
Closed in #1156

PR #757, opened 2018-09-17
Remove probability masking

ansoncfit:
By default, R5 no longer bootstraps estimates of accessibility, which obviates the probability masking features for regional analysis results.  This PR removes masking features and makes a few other small changes:
- Eliminates the fetch of the probability surface from analysis-backend, which was triggering an occasional bug.
- Eliminates the "minimum probability of change" slider
- Corrects cloudwatch logGroup used in link from admin panel
- Changes the default suggestion for regional analysis names (see discussion in #538).  While we're at it, we might also want to tackle #737.
- Refactors calculation of percentiles (steps toward #619, #756)

### DIFF
--- a/lib/components/map/regional.js
+++ b/lib/components/map/regional.js
@@ -18,22 +18,16 @@ type Props = {
   comparisonId: string,
   differenceGrid: Grid,
   isRegionalAnalysisSamplingDistributionComponentOnMap: boolean,
-  minimumImprovementProbability: number,
   probabilityGrid: Grid,
+  removeRegionalAnalysisSamplingDistributionComponentFromMap: () => void,
   setRegionalAnalysisOrigin: Object => void
 }
 
-type DefaultProps = {
-  minimumImprovementProbability: number
-}
 
 /**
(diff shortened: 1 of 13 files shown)


lib/components/map/regional.js
```diff
@@ -45,8 +39,16 @@ export default class Regional extends Component<DefaultProps, Props, void> {
...
   componentWillUnmount () {
+    const {
+      isRegionalAnalysisSamplingDistributionComponentOnMap,
+      removeRegionalAnalysisSamplingDistributionComponentFromMap
+    } = this.props
     const {map} = this.context
     map.off('click', this.onMapClick)
+    if (isRegionalAnalysisSamplingDistributionComponentOnMap) {
+      // don't leave dest travel time distribution on map when exiting analysis mode
+      removeRegionalAnalysisSamplingDistributionComponentFromMap()
```
trevorgerhardt: Map component adding/removing is no longer done this way since map components are now controlled directly within the individual pages.

PR #759, opened 2018-09-19
Fix bounds setting

trevorgerhardt:


lib/components/analysis/regional.js
```diff
@@ -40,11 +42,29 @@ export default class RegionalAnalysis extends React.PureComponent {
...
   }
 
+  _isComplete () {
+    const p = this.props
+    return !!p.analysis && (p.analysis.status == null || p.analysis.status.complete === p.analysis.status.total)
+  }
+
+  _getAnalysisBounds () {
+    const {north, west, width, height, zoom} = this.props.analysis
+    const nw = LeafletMap.prototype.unproject([west + 1, north], zoom)
```
ansoncfit: +1 ?

lib/components/analysis/advanced-settings.js
```diff
@@ -93,18 +92,16 @@ export default class AdvancedSettings extends Component {
...
+    const {profileRequest, regionBounds, regionalAnalyses} = this.props
+    const {bounds} = profileRequest
 
     // figure out which is selected
     let selected
-    if (
-      !regionalAnalysisBounds ||
-      boundsEqual(regionalAnalysisBounds, regionBounds)
-    ) {
+    if (!bounds || boundsEqual(bounds, regionBounds)) {
```
ansoncfit: Where is boundsEqual defined?

lib/components/analysis/regional.js
abyrd: Yeah it's strange to see west + 1 but not north + 1...

trevorgerhardt: Copied code. No comments on the original either. 🤔

Issue #710, opened 2018-07-11
In first analysis for a new project, origin pin location set from previous project


ansoncfit:
If the previous project is in a distant region, a user is unlikely to realize there's a problem (see #337, #604)

PR #761, opened 2018-09-20
Set region data correctly

### LINKED ISSUE #744: Items from one region seen in another region
Lists of opportunity datasets not updated when switching regions

Example:

- https://analysis-staging.conveyal.com/regions/5a9f74f932b98e11405c83dd/analysis: Opportunity dataset dropdown has lodging in Tokyo
- Click on Regions
- Click on Test Region - Boston
- Click on Analyze
- Opportunity dataset dropdown still has Tokyo data.  Refreshing the page loads proper Boston data.

The same issue appears to affect the dropdown options in the opportunity dataset management page.

trevorgerhardt:
This PR fixes #744 and #710 by:

1. Loading Opportunity Datasets each time the selector is shown (on page navigation)
2. Loading Regional Analyses when navigating to the analysis page
3. Setting the entire profile request--instead of just updating it--on region change

### DIFF
--- a/lib/reducers/__tests__/__snapshots__/analysis.js.snap
+++ b/lib/reducers/__tests__/__snapshots__/analysis.js.snap
@@ -431,48 +431,6 @@ Object {
 }
 `;
 
-exports[`reducers > analysis should handle set profile request 1`] = `
-Object {
-  "active": false,
-  "isFetchingIsochrone": false,
-  "profileRequest": Object {
-    "accessModes": "WALK",
-    "bikeSpeed": 4.166666666666667,
-    "date": "2016-12-20",
-    "directModes": "WALK",
-    "egressModes": "WALK",
-    "fromTime": 24321,
-    "maxRides": 4,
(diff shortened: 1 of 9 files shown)


lib/reducers/__tests__/__snapshots__/analysis.js.snap
```diff
@@ -502,29 +511,28 @@ Object {
...
+      "modificationId": "Add Trip Pattern",
+      "title": "Error while applying the modification entitled 'Add Trip Pattern'.",
+    },
+  ],
   "selectedBookmark": null,
 }
 `;
 
-exports[`reducers > analysis should handle set scenario application errors 1`] = `
+exports[`reducers > analysis should handle update profile request 1`] = `
```
ansoncfit: Was the test for showing scenario application errors removed?

trevorgerhardt: Nope, they just have similar output and were shifted up.

Issue #768, opened 2018-09-26
Update single point accessibility numbers on percentile change


trevorgerhardt:
Related to https://github.com/conveyal/analysis-ui/issues/756, the isochrone currently gets updated but the accessibility numbers above the stacked percentile graph do not.

PR #769, opened 2018-09-26
Correct accessibility numbers shown

trevorgerhardt:
Accessibility numbers shown did not reflect the percentile selected and stayed on the 50th even when
changed.

#768


ansoncfit:
Is "median" still displayed in the progress bar?

trevorgerhardt:
Yes, will remove that too.

lib/components/analysis/stacked-percentile-selector.js
```diff
@@ -82,18 +84,7 @@ export default class StackedPercentileSelector extends PureComponent {
...
-          <strong>{opportunityDatasetName || `[select opportunity dataset]`}</strong>
-          {' '}
-          within
-          {' '}
-          <strong>{isochroneCutoff}</strong>
-          {' '}
-          minutes
-        </h5>
-
+        <h5>Accessibility in <strong>{isochroneCutoff}</strong> minutes with <strong>{nearestPercentile}</strong>th percentile reliability</h5>
```
ansoncfit: If there's space, I'd prefer not to lose having the opportunity dataset name shown.  Also, I'd like to move away from "reliability," as it could connote on-time performance, which we don't measure.  I'd suggest:

Accessibility to <strong>{opportunityDatasetName || `[select opportunity dataset]`}</strong> in {isochroneCutoff} minutes (travel time percentile: {nearestPercentile})

lib/components/analysis/stacked-percentile-selector.js
```diff
@@ -120,9 +111,7 @@ export default class StackedPercentileSelector extends PureComponent {
                 width: `${(accessibility || 1) / maxAccessibility * 100}%`
               }}
             >
-              <div className='project-name'>
-                {`${projectName} (Median)`}
-              </div>
+              <div className='project-name'>{projectName}</div>
```
ansoncfit: 👍

PR #770, opened 2018-09-26
Add service start and end dates to bundles

### LINKED ISSUE #478: Analysis date out of GTFS validity
The default value of the date of analysis is "Today". It can be out of the valid range of dates in the GTFS Bundle. The analysis produces bogus results.

The default date should be in the valid range in the GTFS bundle, which could be either calculated or an input in the "configure panel settings" panel.

trevorgerhardt:
Checks service dates in feeds on creation and adds the start and end dates.

This PR also makes a change that directly load bundles and projects instead of retrieving them with the region and updates the actions, selectors, and reducers accordingly.

closes #478
requires https://github.com/conveyal/analysis-backend/pull/171

screenshot 2018-09-27 19 50 47


trevorgerhardt:
closing in favor of simpler changes

configurations/default/messages.yml
```diff
@@ -267,6 +267,7 @@ analysis:
   gisExport: Export to GIS
   downloadRegional: Download %s
   downloadRegionalProbability: Download probability of improvement
+  bundleOutOfDate: Selected date %(selectedDate) is out of service for project %(project). It's bundle %(bundle) only has service between %(serviceStart) and %(serviceEnd).
```
ansoncfit: "Its bundle"

lib/actions/region.js
```diff
@@ -66,13 +67,18 @@ export const load = (id: string, done?: Function) => (dispatch: Function) =>
...
       const region = response.value
+
+      // For legacy regions that were accidentally storing them in the region
+      delete region.bundles
+      delete region.projects
       dispatch([
-        setLocally({...region, projects: []}),
+        setLocally(region),
+        loadBundles({regionId: id}),
+        loadProjects({regionId: id}),
```
ansoncfit: I'm a bit confused on what happens here.  What was saved locally before, and what is saved locally now?

configurations/default/messages.yml
trevorgerhardt: One of those grammatical mistakes I make every day 😅

lib/actions/region.js
trevorgerhardt: Everything was contained within the region at one point due to the way it was retrieved. Since we don't re-filter them out on save, this causes the regions to be saved with all of their projects and bundles.

Now this just pulls bundles up one level like we had previously done with projects.

trevorgerhardt:
Less of a step towards that and more of "removing magic" and making the hierarchy clearer. It's not necessary hence why I initially closed it and was going to redo the PR. But it should be a small enough change so I decided to leave it in.

PR #772, opened 2018-10-02
Fully switch to woonerf/message

### LINKED ISSUE #763: Switch all messages -> message and remove direct sprintf dependency
See https://github.com/conveyal/woonerf#message

trevorgerhardt:
Lot's of surface are here, but nothing should "break" if it's incorrect. Possibly just have an `undefined` show up somewhere that I missed. Snapshot tests do catch a lot of the changes.

Closes #763


evansiroky:
All the code changes in here makes sense.  However, I wouldn't know how to go about making sure all the messages make sense in each branch of the display.  Maybe @ansoncfit can do a quick walk-through to make sure the app is correctly displaying messages in most of the critical places.

PR #773, opened 2018-10-02
Run new linter/prettifier on code in preparation of mastarm update

trevorgerhardt:
Mastarm has been updated significantly and the main reason why I haven't updated it in Analysis UI is due to lint/flow/jest tests.

This is a first step that is mainly `mastarm lint --fix` using the new mastarm but will still pass our current linter on the old mastarm.

### DIFF
--- a/lib/actions/get-feeds-routes-and-stops.js
+++ b/lib/actions/get-feeds-routes-and-stops.js
@@ -2,7 +2,6 @@
 import fetch from '@conveyal/woonerf/fetch'
 import uniqBy from 'lodash/uniqBy'
 
-import {lockUiWithError, setFeeds} from './'
 import {
   ADJUST_DWELL_TIME,
   ADJUST_SPEED,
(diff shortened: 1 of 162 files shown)


lib/actions/get-feeds-routes-and-stops.js
```diff
@@ -12,9 +11,10 @@ import {
   REROUTE
 } from '../constants'
 import * as query from '../graphql/query'
-
 import type {Modification} from '../types'
 
+import {lockUiWithError, setFeeds} from './'
```
evansiroky: Is there a linting rule you're trying to satisfy with placing the imports from the current directory at the very end?

lib/modules/admin/components/job-dashboard.js
```diff
@@ -13,11 +13,11 @@ const formatStartTime = (ms: number) =>
 class JobDashboard extends React.PureComponent {
   props: {
     allBundlesById: Object,
-    jobsWithWorkers: Job[],
-
     clearJob: () => void,
+
```
evansiroky: might make sense to remove this space here

lib/actions/get-feeds-routes-and-stops.js
evansiroky: nvm, I see the pattern that the linter used to fix it.  I think it makes sense.

PR #774, opened 2018-10-02
Show analysis bounds and snap to the grid

### LINKED ISSUE #771: Always show analysis bounds on map
Currently we only show the bounds of the analysis on the map while "set custom geographic bounds" is active. We could just show the bounds all the time (with a non-filled dotted line rectangle or even just the corners). This way there would be no surprise effects when regional analyses were launched (and even on single-points people would visually see if their opportunity counts were being limited to a sub-area). When "bounds of analysis" was set to "same as existing analysis X" the bounds would update, and you would see the bounds rather than waiting until a regional analysis is already running.

trevorgerhardt:
- [x] Show bounds of analysis
- [x] Snap custom bounds to the same grid that we use when running analysis so that the geographic coordinates do not deviate from the binned pixel grid.
- [x] Add a note explaining the above

closes #771


trevorgerhardt:
@ansoncfit do we have a section in the docs about how and why we use a regular grid in analysis? That would be great to link to for any explanation of why we're snapping the bounds here.

ansoncfit:
I just expanded that section (https://github.com/conveyal/analysis-ui/pull/777).  Link should be
http://docs.analysis.conveyal.com/en/latest/analysis/methodology.html#spatial-resolution

Issue #734, opened 2018-08-08
Add confirmation alert to Fetch LODES


ansoncfit:
Are you sure you want to fetch updated LODES data?  These data are only available for the United States

Issue #718, opened 2018-07-27
Missing mode selectors


ansoncfit:
⛴  FERRY
🚃  CABLE_CAR
🚡  GONDOLA
🚞  FUNICULAR

Mode TRANSIT is no longer sent (after #489; and it wouldn't work in FastRaptorWorker @ 224), so the these modes are excluded from analysis with no way for users to turn them on.  We need ferries for Boston (and probably Gondolas for Portland). I'd be fine with all of these being grouped into one button for "Other"

trevorgerhardt:
It may be just as easy to add all of the buttons as we may get subsequent requests to split them out (for a place like Medellin). Thoughts on this @abyrd?

PR #776, opened 2018-10-06
More buttons

ansoncfit:
Fixes for #734 and #718.

Tests still need to be updated to reflect the default profile request now including transit modes = "BUS,TRAM,RAIL,SUBWAY,FERRY,CABLE_CAR,GONDOLA,FUNICULAR"

The new transit mode selector could be refactored for reuse in add-trip modification details, to set the mode of new routes.  I thought we had a ticket open for that, but I can't find it.


lib/components/analysis/mode-selector.js
```diff
@@ -25,7 +25,12 @@ const BUS = 'BUS'
 const RAIL = 'RAIL'
 const TRAM = 'TRAM'
 const SUBWAY = 'SUBWAY'
-const ALL_TRANSIT = [BUS, RAIL, TRAM, SUBWAY].join(',')
+const FERRY = 'FERRY'
+const CABLE_CAR = 'CABLE_CAR'
+const GONDOLA = 'GONDOLA'
+const FUNICULAR = 'FUNICULAR'
+const ALL = 'ALL'
+const ALL_TRANSIT = [BUS, RAIL, TRAM, SUBWAY, FERRY, CABLE_CAR, GONDOLA, FUNICULAR].join(',')
```
trevorgerhardt: It's better to just have two variables here like `ALL_TRANSIT` and `ALL_TRANSIT_STRING` instead of splitting the string again later one.

lib/components/analysis/mode-selector.js
```diff
@@ -35,32 +40,46 @@ export default class ModeSelector extends Component {
...
-      transitModes
+      directModes: newMode
     })
   })
 
   _toggleTransitMode = memoize(mode => () => {
-    const transitModes = this._hasTransit(mode)
-      ? without(this.props.transitModes.split(','), mode).join(',')
-      : [this.props.transitModes, mode].filter(Boolean).join(',')
+    var transitModes
```
trevorgerhardt: Use `let` instead of `var`

PR #795, opened 2018-10-23
Clear out of date isochrones

### LINKED ISSUE #617: Gray Opportunity Plot When Stale
Use selectProfileRequestHasChanged to also make the cumulative opportunity plot grayscale when the isochrone is stale.

### LINKED ISSUE #684: Disable creation of regional analyses unless isochrones are displayed
We don't have a way of reporting errors that arise handling regional tasks (e.g. errors with scenario application).  

We should force users to "pre-flight" by performing a single-point analysis before running a regional analysis.  Unless an isochrone is displayed, the button to create a new regional analysis could be disabled, with a tooltip instructing users "Fetch results for a single point before running a regional analysis."  Forcing users to check a single-point request would give them the chance to see any scenario application errors and could also help reduce instances of them starting analyses with the wrong settings (e.g. dates with no transit service).

### LINKED ISSUE #794: Don't show out-of-date isochrones
Trying to reproduce #687, I fetched isochrones, modified an add-trip modification by moving the second stop from West Medford to East Arlington, navigated back to the analysis panel, and saw this:
image

Once I fetched results, the results were as expected.  So this isn't unexpected behavior from our perspective, but it could be quite misleading for users.

We shouldn't show isochrones that don't correspond to the most up-to-date scenarios.  Suggestions:
1. Remove isochrones when a selected/comparison scenario is modified.  Or at least remove them when navigating away from the analysis page.
1. Disallow adjustment of travel time and percentile sliders (as well as isochrone download) when users select different projects/scenarios or when profile request settings are changed (i.e. when isochrones are grey).  I've seen multiple users adjust sliders after changing scenarios but before fetching results, which leads me to believe they think the isochrones have actually been updated.

trevorgerhardt:
This change 
- disables most inputs when a project isn't selected
- disables most inputs when an isochrone is fetching
- removes isochrones that are not current with the selected inputs
- grays out the stacked percentile selector when stale (could disable it?)
- only allows creating a regional analysis when an isochrone has been fetched

closes #794
closes #684 
closes #617


lib/components/analysis/index.js
```diff
@@ -304,22 +310,18 @@ export default class SinglePointAnalysis extends Pure {
 
             <ModificationsMap />
 
-            {isochrone &&
+            {!profileRequestHasChanged && !disableInputs && isochrone &&
```
ansoncfit: For future maintainability, comments explaining these different booleans might be helpful. For example, does changing the selected project or scenario set `profileRequestHasChanged` to true?

trevorgerhardt: Anything that affects the requested data and not just the displayed data. Disabling inputs occurs on fetch and when there are fetch errors, so it also seemed appropriate here.

lib/components/analysis/index.js
```diff
@@ -495,7 +498,7 @@ export default class SinglePointAnalysis extends Pure {
                 * point mode.
                 */}
               <Slider
-                disabled={disableInputs}
+                disabled={profileRequestHasChanged}
```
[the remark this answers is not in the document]
trevorgerhardt: True 👌 

The removing of the isochrones on change seemed more straightforward, but I had made this change first.

trevorgerhardt:
@ansoncfit follow up note: the whole analysis form is convoluted right now with the cumulative opportunity data and results smack in the center making it a mix of "requested inputs", "results", and "displayed data inputs". 

So while it's not perfect, unless there are errors or I missed something this should be astep in the right direction but could use more future work.

Also, showing grayed out isochrones but not changing them is based on the data inputs is possible but not advisable by me with the current state of things.

ansoncfit:
Agreed, it's a useful step in the right direction, and we can do a more thorough redesign of the analysis form later.

I'm worried users will perceive this as broken if the Regional Analysis button is disabled without explanation.  What did you think of the suggested tooltip in #684?  It might also be worth adding a tooltip to fetch results when that is disabled -- "Select a project and scenario before fetching results".  That same tooltip could be added to the origin marker, which is also disabled when a project/scenario have not yet been set.

ansoncfit:
Also closes #617, right?

trevorgerhardt:
Added tooltips and dimmed the marker 👌

PR #800, opened 2018-10-30
Abort fetching isochrones

trevorgerhardt:
- Add an abort button when fetching an isochrone. 
- Auto abort when you leave the page. 
- Increase the timeout retry delay to 30 seconds and limit it to ten retries.

NOTE: This includes the latest `woonerf` which removes duplicate data from the `decrement fetch` action so a lot of snapshots have been updated.

### DIFF
--- a/lib/actions/analysis/index.js
+++ b/lib/actions/analysis/index.js
@@ -5,11 +5,12 @@
  */
 
 import lonlat from '@conveyal/lonlat'
-import {fetchError, fetchMultiple} from '@conveyal/woonerf/fetch'
+import {fetchMultiple} from '@conveyal/woonerf/fetch'
 import {createAction} from 'redux-actions'
 
 import {
   ANALYSIS_URL,
+  FETCH_TRAVEL_TIME_SURFACE,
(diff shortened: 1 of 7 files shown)


lib/actions/analysis/index.js
```diff
@@ -27,7 +28,8 @@ import type {GetState, LonLat} from '../../types'
 
 import {parseTimesData} from './parse-times-data'
 
-const RETRY_TIMEOUT_MILLSECONDS = 5000
+const RETRY_TIMEOUT_MILLISECONDS = 1000 * 30 // 30 seconds
```
ansoncfit: 👍

trevorgerhardt:
Ah, JavaScript fun. Passing an action directly as an event handler and it passes the event and I was just checking existence of a variable not if it was `true`.

trevorgerhardt:
I had also been debugging by moving the marker.

trevorgerhardt:
Ah, yes, using those tooltips for buttons was not that straightforward due to the wrapper that's added. I need to figure out a better solution but for now I think it'd be best if we just go through with the tooltips we have, even though they are delayed.

Issue #746, opened 2018-08-22
Make "Baseline" scenario available in dropdown


ansoncfit:
It's available for the "Comparison scenario", and it would be hepful to have as an option for the main scenario as well.

Issue #255, opened 2017-01-27
GIS export of scenario


mattwigway:
I've had a number of requests to export the scenario to GIS. I think that the output most people are looking for here is the GIS of every line in the feed including modifications. This is non-trivial but not impossible; for reroutes, we'd generate new shapes, add-trips is easy, and remove stops/trips shouldn't be too too hard except for truncation of shapes with remove-stops.

ansoncfit:
As a first step, the ability to export .geojson of the alignments of add-trip modifications would be helpful.

ansoncfit:
Closing until we have explicit request for export showing removed stops, segments with altered speeds, etc.

PR #803, opened 2018-11-06
Scenario management

ansoncfit:
In the current interface, the hierarchy of scenarios and modifications is confusing.  The modifications list title has the project name, then "Modifications", then a list of scenarios, then options for creating modifications.  This PR:
* Addresses that unclear flow using a collapsible scenario list

image

* Reduces visual clutter, in part by combining the infrequently used export/reporting options in the project title
* Makes it clearer what the (previously hidden) Baseline scenario is.
* - [ ]  #746
* - [ ] initial pass at part of #255

### DIFF
--- a/configurations/default/messages.yml
+++ b/configurations/default/messages.yml
@@ -143,6 +143,7 @@ project:
   select: Select a project…
 variant:
   name: Scenario
+  baseline: Baseline
   plural: Scenarios
   enterName: Enter a name for your scenario:
   editName: Rename this scenario
--- a/configurations/default/messages.yml
+++ b/configurations/default/messages.yml
@@ -134,7 +134,7 @@ project:
   selectBundle: Select a GTFS bundle...
   noBundles: To get started with projects in this region, first create a GTFS bundle.
   print: Print a scenario from this project
-  export: Download a scenario from this project
+  export: Download/share a scenario from this project
   route: Route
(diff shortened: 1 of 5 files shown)


trevorgerhardt:
Rearranging it to be above modifications is a good idea, really an artifact of the old design. We should probably make it look similar to modifications right below it...if that's the case then create should be a button also.

ansoncfit:
Thanks for the suggestion, @trevorgerhardt.  

Snapshots still need updating, but this should be ready for review now.  I could use some help on the CSS to prevent the expanded scenario list from making the whole window scrollable:

image

trevorgerhardt:
@ansoncfit because of the way we have the scrolling set up for the dock that wouldn't be the easiest to implement. How do you feel about this kind of look?

screenshot 2018-11-07 12 46 39
screenshot 2018-11-07 12 47 34

ansoncfit:
I'd be in favor of it as a good incremental improvement.

trevorgerhardt:
Shouldn't that download button go next to `Scenarios` now? The language seems much more directly tied to them now.

ansoncfit:
Yeah, that would make sense.  I didn't get around to dealing with the `block` of the collapsible title.

Feel free to take over and add commits to this branch.

trevorgerhardt:
Cool, will do!

Issue #805, opened 2018-11-07
Download stops of add-trip modifications


ansoncfit:
As geojson, including auto-generated stops

Another step toward #255

Issue #667, opened 2018-04-03
Reused vs. new stops less visible when editing routes


abyrd:
The new route alignment editing looks great but it's now harder to tell the difference between new and existing (reused from GTFS) stops when creating routes. Reusing stops can give a significant speedup over creating new ones, so it’s very helpful to highlight the difference. Often there are already well-positioned stops that can be reused and the visual cue helps remember to reuse them. I used to almost find it almost like a game to see how many solid blue dots I could get when drawing a route.

trevorgerhardt:
Reused stops are black instead of blue. "this already exists" vs "this is new". What alternate highlighting do you suggest?

abyrd:
I think they used to be solid vs. filled. At least on my screen / in my lighting conditions / etc. when I saw the new symbology (which is great to have visible right in the app) I could no longer readily see the difference between them.

abyrd:
I just took another look at it. I see there are already filled blue stops. Another symbol that might provide a more obvious contrast would be filled gray. The current system heavily emphasizes control points because they're filled blue. A circle with a plus in it might be good for new (non-reused) stops, and a square or diamond as a non-stop control handle.

ansoncfit:
+1 for diamond control points.

Nate-Wessel:
I went down a little rabbit hole and designed a better set of symbols. 

![new stop legend](https://user-images.githubusercontent.com/7830465/72475269-94e88d00-37b8-11ea-9c50-3778ac42d239.png)

* The original stop is a smaller grey circle. 
* A reused stop is a larger, darker circle around the same symbol. (i.e. visually selecting it and pulling it forward, and raising it to the foreground)
 * The control point stays a filled circle, but with a thin darker blue border to better distinguish it from the line. 
* New stops contain a + sign, which stands out pretty well visually. 
* Automatic spacing stops have a - sign and a grey rather than black border. (maybe not the best solution, but I tried a few other things that didn't work well.)

All symbols have a somewhat thicker border than the current symbols. which helps them to stand out better from the background. These are better differentiated than the current set, using color, size and shape to distinguish between them.

Issue #654, opened 2018-03-05
Allow users to reverse stop order of add-trip and re-route modifications


ansoncfit:
Data Tools has a handy reverse trip pattern feature:
![image](https://user-images.githubusercontent.com/2173529/36993429-ca550d70-207b-11e8-9613-50ffd6dbf4ef.png)

In add-trip modifications, users might want a basically bidirectional route, but with small variations.  It would be nice to draw one direction, copy it, reverse the stop order, then make the small changes.

For re-route modifications that represent extensions, users usually want a bidirectional extension.  But they currently have to draw out stops twice, once per pattern/direction.

Nate-Wessel:
This is also helpful if you draw a bidirectional line and later decide to split it into two directions, for example to phase it with another line. 
Could also help with #949. The reroute geometries do not seem to display a direction property, though they seem to have it when you try to edit them.

ansoncfit:
For add-trip modifications, 
- segments, segmentSpeeds, and dwellTimes arrays need to be reversed
- segments.stopAt{Start/End} need to be switched
- segments.{from/to}StopId need to be switched
- segments.geometry.coordinates array needs to be reversed

Something like:
```
newSegmentSpeeds = segmentSpeeds.reverse()
newDwellTimes = dwellTimes.reverse()

newSegments = []

for(const s of segments.reverse()){
    const stopAtStart = s.stopAtEnd
    const stopAtEnd = s.stopAtStart
    const fromStopId = s.toStopId
    const toStopId = s.fromStopId
    const spacing = s.spacing
    const geometry = s.geometry
    geometry.coordinates.reverse()
    newSegments.push({
        stopAtStart,
        stopAtEnd,
        fromStopId,
        toStopId,
        spacing,
        geometry    
    })
}
```

ansoncfit:
Just discussed as a client priority.

PR #809, opened 2018-11-08
Alignment Editing Enhancements

### LINKED ISSUE #657: Change "Extend from end" in route editor to be trinary
"Extend from: start, end, neither"

Users may need to edit a route geometry but have no intention of extending it.  Being able to click around without spuriously extending would be nice.  This change would also help #646 from being too obtrusive

ansoncfit:
Added feature: Snap a reroute/add-trips alignment to nearby stops from the baseline network GTFS. Stops must be within `STOP_SEARCH_BUFFER_KM` of the alignment to be considered for snapping.  To prevent extraneous stops from being added (e.g. stops in a pair across the street from one another, or four stops at an intersection), stops are clustered using turf's dbscan, and only one stop (the one closest to the alignment) in each cluster is added.  `APPROX_MIN_SPACING_KM` is used as the clustering radius, and to prevent stops from being added too close to stops that are already explicitly specified in the modification.

Follow streets is not yet implemented -- it would be substantially messier and more difficult, as it might generate multiple unintended U-turns if snapped stops were on opposite sides of the street, etc.

Also fixes #657 

Leaving for future work:
- [ ] #805
- [ ] #667
- [ ] #654 
- [ ] #480


lib/containers/reroute.js
```diff
@@ -3,6 +3,7 @@ import {connect} from 'react-redux'
 
 import Reroute from '../components/modification/reroute'
 import * as select from '../selectors'
+import getStopsFromFeeds from '../utils/get-stops-from-feeds'
```
trevorgerhardt: Hmmm...lets turn this into a selector, that way it doesn't have to be re-run if the feed doesn't change.

trevorgerhardt:
Unless you've already made progress on those other five tasks I'd recommend keeping the scope down in this PR and working on those in subsequent PRs 👌

lib/selectors/existing-stops-along-pattern.js
```diff
@@ -0,0 +1,104 @@
...
+// TODO use async function that allows following streets
+// import getLineString from './get-line-string'
+const createLineString = (from, to) => {
+  return {type: 'LineString', coordinates: [from, to]}
+}
+
+const STOP_SEARCH_BUFFER_KM = 0.02 // 20 meters on either side of the alignment
+const APPROX_MIN_SPACING_KM = 0.08 // 80 meters
+
+export default function getExistingStopsAlongPattern (segments: Segment[], stops: GTFSStop[]): Segment [] | void {
```
trevorgerhardt: My main request with this PR is to move this file into `utils` or change where it is used to be in a `container` and making it an actual selector with reselect's `createSelector`. Selectors are a standard way of computing and caching derived data from state.

trevorgerhardt: Selectors can be used together: https://github.com/conveyal/analysis-ui/blob/dev/lib/selectors/isochrone.js#L65

And the standard style allows them to be used simply like: https://github.com/conveyal/analysis-ui/blob/dev/lib/containers/single-point-analysis.js#L47

lib/containers/reroute.js
ansoncfit: I'm realizing now that I misread this comment when I first saw it.  Instead of refactoring the existing `getStopFromFeeds` util to a selector, I started to move the new `existing-stops-along-pattern` function, which you later pointed out didn't make sense.  Sorry for the confusion.

Now starting the suggested refactor (see https://github.com/conveyal/analysis-ui/commit/dc83b0e77e41d4f73b412c63fae1d329564928d0)

PR #817, opened 2018-11-14
v4.5.0

trevorgerhardt:
Features:
* Customize JSON directly in Advanced Settings
* Show new messages returned from worker when it is starting up
* Parse GeoTIFFs when their `Content-Type` is not `application/octet-stream`

Bug Fixes:
* Create bookmarks with a comparison project selected
* Properly handle timeouts when a retry is no longer occuring


ansoncfit:
Will #818 make it into this release?

Heads up @evansiroky -- an upcoming merge to master, for which we may want to consider additional tests for coverage.

abyrd:
Looks good to me so far, though I haven't tested it on staging. Hopefully I will do that today before/during our work session and then we can merge if everything is operating smoothly.

PR #818, opened 2018-11-14
Address nonce issues

trevorgerhardt:
Stacked saves from editing alignments should not conflict with returned nonces anymore. Nonce error throwing behavior is inconsistent so I'm not 100% certain this fixes it but should be a step in the right direction.

Also, fix copy timetables that was broken when we updated woonerf.

### DIFF
--- a/lib/components/modification/editor.js
+++ b/lib/components/modification/editor.js
@@ -1,6 +1,7 @@
 // @flow
 import Icon from '@conveyal/woonerf/components/icon'
 import message from '@conveyal/woonerf/message'
+import debounce from 'lodash/debounce'
 import React from 'react'
 
 import {Application, Dock, Title} from '../base'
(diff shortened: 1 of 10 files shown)


lib/components/modification/editor.js
```diff
@@ -62,29 +63,14 @@ export default class ModificationEditor extends React.Component {
...
   componentDidMount () {
-    window.addEventListener('beforeunload', this._checkForUnsavedChanges)
     this.props.setActive()
-    this._savedChangesInterval = setInterval(this._saveChanges, AUTOSAVE_EVERY_MS)
   }
 
   componentWillUnmount () {
-    window.removeEventListener('beforeunload', this._checkForUnsavedChanges)
-    clearInterval(this._savedChangesInterval)
-    this._saveChanges()
```
ansoncfit: Are changes still saved immediately when navigating away from the modification?

trevorgerhardt: Not immediately, I can do that with the `flush` function though so I'll add that back.

trevorgerhardt: Hmmm, I've figured out a way to trick it with `flush`. Back to the drawing board.

Issue #820, opened 2018-11-19
Routing errors not displayed properly


ansoncfit:
When handling analysis requests, we can have scenario errors or routing errors.  The former results in an array of errors, which show up in the side panel.  The latter may include a stacktrace, but this results in a front-end error:

```Uncaught (in promise) TypeError: Cannot read property 'map' of undefined
    at ScenarioApplicationError.render (scenario-application-error.js:30)```

To reproduce, set walkSpeed = 0.  If memory serves me correctly, such routing errors used to result in a modal with the stacktrace.

PR #821, opened 2018-11-19
Show errors

ansoncfit:
Fixes two problems:
1. Scenario application warnings were not being shown (should be in side panel)
1. Miscellaneous routing errors and AnalysisServerExceptions were not being shown (should be in modal, #820)

### DIFF
--- a/lib/actions/analysis/index.js
+++ b/lib/actions/analysis/index.js
@@ -201,7 +201,8 @@ const handleGeoTIFF = (names) => (responses) => {
  * Handle response for a travel time surface request.
  */
 export const handleSurface = (profileRequest: any) => (error: any, responses: any) => {
-  if (error) {
+  if (responses.statusCode >= 400 || responses.some(r => r.statusCode >= 400) ||
+    error) {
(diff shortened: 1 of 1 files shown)


trevorgerhardt:
Ok, a few things to make this more robust. 

1. Is this backwards compatible with older versions of R5? 
2. We have "warnings" and "errors". Shouldn't this be in the "error" category? If so, why are we calling it a "warning".
3. Instead of checking for the "existence" of errors we should really be checking the status code. Get the status and check for `>= 400`, then set the errors accordingly.

ansoncfit:
With the latest R5 build (`v4.2.0-106-gbb88602`),
- `walkSpeed = 0` ➡️ ErrorModal with stacktrace
- Add-trip modification with missing timetable ➡️ red error box with collapsible error details in analysis panel

In terms of backwards compatibility, this makes things no worse for R5 v4.1.1:
- `walkSpeed = 0` ➡️ ErrorModal (no stacktrace)
- Add-trip modification with missing timetable ➡️ warning box with collapsible error details in analysis panel

PR #837, opened 2018-12-20
GeoJSON for added features

ansoncfit:
Step toward #255
Download LineString geojson for new segments of add-trip and reroute modifications
Download Point geojson for stops added in add-trip and reroute modifications


ansoncfit:
Closes #805

Issue #833, opened 2018-12-10
Toggle map display does not work for Baseline scenario


ansoncfit:
Clicking 👁  next to scenario 0. Baseline should hide all modifications from the map, but it appears not to.

ansoncfit:
I left this as a TODO at https://github.com/conveyal/analysis-ui/blob/968b0b993e95b4297c30f7496176e8b5b5f76e40/lib/components/variant-editor.js#L72

Issue #807, opened 2018-11-07
Titles of accessibility statistic progress bars undefined/misleading


ansoncfit:
While working on #409, we should also clean up these titles:

![image](https://user-images.githubusercontent.com/2173529/48164160-b8a86d00-e2af-11e8-9cc0-369609fa9fcd.png)

- [ ] undefined scenario should be "Baseline" (something I overlooked in c34073690ae1ffdb7a4c6fd5e908d36a8b9bd108)
- [ ] full project and scenario names should be shown.  The full name of the second scenario is "Connector Removed", so it is potentially misleading/confusing for the scenario title to be cropped to "Connector" when the progress bar is low.

PR #838, opened 2018-12-31
Project/scenario names

ansoncfit:
This tries to make analysis panel/map behavior more intuitive when displayed accessibility statistics/isochrones are stale (i.e. the current profile request settings or selected project/scenario differ from the settings of the displayed analysis). The main goal is to let users see stale results so they can see changes easily when results update, but to disallow them from downloading or interacting with stale results.

More specifically, this PR:
- Hides download isochrone geojson/geotiff buttons unless an isochrone is showing, and disables these buttons if isochrones are stale
- Reverts to showing stale isochrones (in gray), but disables the travel time and percentile sliders when isochrones are stale
- Adds `comparisonVariant` and `comparisonProject` to profile requests, so the `profileRequestHasChanged` selector will return false if the comparison project/scenario differ from what are displayed. This changes includes new selectors for displayed project/scenario names. We may also want to consider a larger refactor, in which we use an array of different profile requests to facilitate other settings differing between the main and comparison isochrones.
- Uses the full project-scenario names from the displayed profile request (rather than the current one) for the stacked percentile chart and accessibility progress bars, resolving #807. If a user sets a different project/scenario, the main accessibility title will change to "Accessibility will update after fetching results", the bars will go grayscale, and the project/scenario titles will not update, until the corresponding results are actually fetched and displayed.
- Includes small fix for #833


lib/components/analysis/map.js
```diff
@@ -43,9 +43,6 @@ const isochroneStyle = (fillColor) => ({
   stroke: false
 })
 
-const mainIsochroneStyle = isochroneStyle(colors.PROJECT_ISOCHRONE_COLOR)
-const compIsochroneStyle = isochroneStyle(colors.COMPARISON_ISOCHRONE_COLOR)
```
trevorgerhardt: Keep these

trevorgerhardt: (and see below)

trevorgerhardt: and add a `const staleIsochroneStlye = ...`

lib/components/analysis/map.js
```diff
@@ -108,18 +107,22 @@ export default class AnalysisMap extends React.Component {
 
         <ModificationsMap />
 
-        {p.displayedDataIsCurrent && p.isochrone &&
+        {p.isochrone &&
           <GeoJSON
             data={p.isochrone}
             key={s.mainIsochroneKey}
-            style={mainIsochroneStyle}
+            style={isochroneStyle(p.displayedDataIsCurrent
```
trevorgerhardt: Switch this to `style={p.displayedDataIsCurrent ? mainIsochroneStyle : staleIsochroneStyle}`

trevorgerhardt: Since `isochroneStyle` returns an object the `===` comparison would be false every time this re-renders.

lib/containers/__tests__/__snapshots__/single-point-analysis.js.snap
```diff
@@ -2675,34 +2675,22 @@ exports[`Containers > Single Point should render correctly 1`] = `
...
                     comparisonInProgress={false}
                     comparisonPercentileCurves={null}
-                    comparisonProjectName={null}
+                    comparisonProjectName="Undefined"
                     disabled={false}
                     isochroneCutoff={60}
                     maxAccessibility={-Infinity}
                     nearestPercentile={50}
-                    projectName="Mock Project: Default"
+                    projectName="Undefined"
```
trevorgerhardt: Is this the intended result?

lib/components/analysis/map.js
ansoncfit: 👍

lib/utils/clean-project-scenario-name.js
```diff
@@ -1,12 +1,14 @@
...
+import {UNDEFINED_PROJECT_NAME} from '../constants'
+import message from '@conveyal/woonerf/message'
 
 export default function cleanProjectScenarioName (
   project?: Project,
   scenarioIndex: number
 ): string {
-  if (!project) { return 'Undefined' }
-  const scenarioName = project.variants[scenarioIndex] || 'Baseline'
+  if (!project) { return UNDEFINED_PROJECT_NAME }
```
ansoncfit: @trevorgerhardt using "Undefined" feels a bit sloppy, but hopefully it's safe enough with this constant?

lib/containers/__tests__/__snapshots__/single-point-analysis.js.snap
ansoncfit: Yes - `StackedPercentileSelector` now gets its project/scenario names from `displayedProfileRequest`.  Before results are fetched and displayed, these will be undefined (even if Mock Project: Default is selected).

ansoncfit: * will be "Undefined".  See comment below.

lib/utils/clean-project-scenario-name.js
trevorgerhardt: I would just avoid "Undefined" because of its counterpart in code, `undefined`. I'd go with "No name" or "(unnamed)" or something like that.

trevorgerhardt: @ansoncfit

PR #839, opened 2019-01-06
Fix phasing

### LINKED ISSUE #829: Phase from stop menu populated with stops from wrong pattern
For example, https://analysis-staging.conveyal.com/regions/5a2e9df732b98e221fef89eb/projects/5bff25fc32b98e3794a6f6e2/modifications/5c0773a032b98e3794a6fc2b

If phase from timetable is set to one of the patterns from a Route 71 convert frequency modification, the phase from stops are from Route 1.  Interestingly, if the phase from timetable is set to one of this project's added trips, the correct stops appear.

ansoncfit:
Closes #829


lib/selectors/all-phase-from-timetable-stops.js
```diff
@@ -32,25 +32,39 @@ const selectPhaseModifications = createSelector(
       : {})
 )
 
+/**
+ * Get stops of the trips in a referenced (not necessarily active) modification
+ */
+const selectStopsFromTrip = (state, m) => {
```
ansoncfit: @trevorgerhardt should we make this a selector?

lib/selectors/all-phase-from-timetable-stops.js
```diff
@@ -32,28 +32,41 @@ const selectPhaseModifications = createSelector(
...
       if (!modification) return
-      if (modification.entries) {
+      if (modification.entries) { // get stops from a referenced adjust-frequency mod (which has entries)
         const entry = modification.entries.find(e => e._id === key.split(':')[1])
-        stopsForTimetable[key] = routeStops(state, {
-          modification: {
+        const feedId = modification.feed
+        const feed = feeds[feedId]
+        if (!feed) return
+        stopsForTimetable[key] = getStopsFromTrip(
```
trevorgerhardt: This function and it's parameters are getting a bit confusing. Let's inline it to:

```js
const modificationPatterns = feed.routes.find(r => r.route_id === modification.routes[0]).patterns
const sourceTripStops = modificationPatterns.find(patternTripsInModification([entry.sourceTrip])).stops
stopsForTimetable[key] = sourceTripStops.map(s => s.stop_id)
  .map(getStopFromFeed(feed.stopsById))
  .map(s => ({...s, stop_id: `${feedId}:${s.stop_id}`}))
```

Assuming that's correct, could use the same comments as the function it is replacing.

PR #840, opened 2019-01-15
Clarify default time table dwell time description and make changes to it consistent

### LINKED ISSUE #823: Setting full-trip dwell does not reset per-stop dwells
If I create a timetable in an add trips modification, then set the dwell times on individual stops to different numbers e.g. 0 and 5 minutes, the overall "Dwell time" field just says 0.
If the per-stop dwells are collapsed, it's not obvious whether they exist, nor whether the full-trip dwell or the per-stop dwells are being used.

If I type something in the full-trip dwell field it changes one of the the per-stop dwells, but not all of them, and it seems to change a different per-stop dwell each time (I've seen it change the first or third one).

trevorgerhardt:
screen shot 2019-01-15 at 10 34 31 pm

closes #823


trevorgerhardt:
Good call. I think with that text `00:40 (default)` in the input we no longer need the description I added. 

screenshot 2019-01-16 14 20 50

What do you think @abyrd?

Issue #845, opened 2019-02-13
Create custom modification via JSON


ansoncfit:
The `saveCustomProfileRequest` feature is useful for experimentation. It would be helpful to have something similar to experiment with modifications.  On the front-end, this could be a new modification type (`custom`). I envision the detail pane for this modification type would consist of:
* A `TextArea` for specifying an array of modifications
* The current `Variants` checkboxes for (de)activating modifications in scenarios

We may also want logic to prevent modifications that are children of a `custom` modification from showing up in the list of modifications that are editable via the existing UI.

PR #852, opened 2019-03-14
Add custom modification type

trevorgerhardt:
Will just have a freeform text field and variant editor. Custom types are created with an empty `r5type` parameter that can be used on the R5 side to convert it into corresponding types. This should be useful for testing modifications by updating r5 without needing to update the UI or backend.

TODO:

- [x] Show new modification types in the type list
- [x] Have a catch-all for any new modification types

re #845

### DIFF
--- a/lib/constants/index.js
+++ b/lib/constants/index.js
@@ -24,6 +24,7 @@ export const ADD_TRIP_PATTERN = 'add-trip-pattern'
 export const ADJUST_DWELL_TIME = 'adjust-dwell-time'
 export const ADJUST_SPEED = 'adjust-speed'
 export const CONVERT_TO_FREQUENCY = 'convert-to-frequency'
+export const CUSTOM_MODIFICATION = 'custom'
 export const REMOVE_STOPS = 'remove-stops'
 export const REMOVE_TRIPS = 'remove-trips'
 export const REROUTE = 'reroute'
@@ -34,7 +35,8 @@ export const MODIFICATION_TYPES = [
   CONVERT_TO_FREQUENCY,
   REMOVE_STOPS,
   REMOVE_TRIPS,
-  REROUTE
+  REROUTE,
+  CUSTOM_MODIFICATION
 ]
 
 /**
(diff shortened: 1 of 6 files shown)


lib/constants/index.js
```diff
@@ -34,7 +35,8 @@ export const MODIFICATION_TYPES = [
   CONVERT_TO_FREQUENCY,
   REMOVE_STOPS,
   REMOVE_TRIPS,
-  REROUTE
+  REROUTE,
+  CUSTOM_MODIFICATION
```
abyrd: Only this value is suffixed with `_MODIFICATION`. Is `CUSTOM` a reserved word or already used as an identifier? If not for uniformity maybe it should just be called `CUSTOM`.

lib/constants/index.js
```diff
@@ -24,6 +24,7 @@ export const ADD_TRIP_PATTERN = 'add-trip-pattern'
 export const ADJUST_DWELL_TIME = 'adjust-dwell-time'
 export const ADJUST_SPEED = 'adjust-speed'
 export const CONVERT_TO_FREQUENCY = 'convert-to-frequency'
+export const CUSTOM_MODIFICATION = 'custom'
```
abyrd: Is this the string that ends up in the `type` field of the modification? Is the `type` field of the produced modification modifiable or must it remain `custom`?

lib/constants/index.js
trevorgerhardt: That is just the `const`ant name. The others are more unique so it seemed right to just append `MODIFICATION` to `CUSTOM`.

lib/constants/index.js
trevorgerhardt: Yes. But I can remove that restriction.

trevorgerhardt:
> If I understand correctly, since all modification types now include an advanced free text editor, this creates a type that simply has no other controls than the free text editor. I am wondering if this is geared toward wrapped modifications like:
> 
> ```
> {
>   type: custom,
>   data: {
>     type: congestion,
>     polygon-source: abcd.geojson
>   }
> }
> ```
> or unwrapped modifications like:
> 
> ```
> {
>   type: congestion,
>   polygon-source: abcd.geojson
> }
> ```

What's your preference? @abyrd

PR #853, opened 2019-03-18
v4.7.0

trevorgerhardt:
One minor feature (editing modifications JSON) and two bug fixes. The surface area of these changes is very small so we should be able to deploy to production easily.


abyrd:
@trevorgerhardt should we try to include #852 in this release? I guess #852 (custom modifications) does rely on backend changes that are not yet in production.

trevorgerhardt:
@abyrd we could, but yea it would take a bit more testing and a backend deployment

PR #854, opened 2019-03-25
Use the previous speed instead of a default speed when expanding the segments array

trevorgerhardt:


lib/components/modification/segment-speeds.js
```diff
@@ -75,9 +75,14 @@ export default class SegmentSpeeds extends PureComponent<void, Props, State> {
   _setSegmentSpeedsLengthTo ({segmentSpeeds, segmentDistances, update}: Props) {
     const newLength = segmentDistances.length
     const newSegmentSpeeds = segmentSpeeds.slice(0, newLength)
-    while (newSegmentSpeeds.length < newLength) {
-      newSegmentSpeeds.push(segmentSpeeds[0] || DEFAULT_SEGMENT_SPEED)
+
+    // Add new segment speeds using the previous segment speed or the default
+    // speed if the previous segment speed is not valid. Usually this will only
```
ansoncfit: i.e. We expect the previous segment speed to be invalid when creating the initial segment?

lib/components/modifications-map/transit-editor/index.js
```diff
@@ -280,11 +286,46 @@ export default class TransitEditor extends LayerGroup {
...
+            ss.unshift(ss[0] || DEFAULT_SEGMENT_SPEED)
+          }
+
+          return this._extendSegmentSpeedsTo(ss, segments.length)
+        }
+
+        if (p.modification.type === ADD_TRIP_PATTERN) {
+          const tts = p.modification.timetables
+          if (tts && tts.length > 0) {
+            updateModification({
```
ansoncfit: Calling update conditional on tts.length > 0 implies users will be unable to draw a route unless a timetable has been added. Maybe we should automatically add a default timetable when creating an add-trip modification.

trevorgerhardt: That check was not necessary as the timetables array always exists, but starts at zero length. Since an Add Trip modification is not valid without a timetable I would be in favor of adding a default timetable on creation.

trevorgerhardt: Removed the check, good catch 👍

trevorgerhardt: Since we have the ability to "copy timetable"s if that is a common pattern then adding a default timetable would require an extra deletion. Let's leave it without a default time table for now.

PR #868, opened 2019-04-22
Nextify

trevorgerhardt:
Might be the largest PR ever for Analysis UI. I will expound below. A summary of the major changes are:
1. No longer uses Mastarm / Woonerf
2. Site is now built with [Next](https://nextjs.org/) 
3. Deployed with [Now](https://zeit.co/docs)
4. Authenticated via dedicated site
5. Stripped all `flow` types
6. Formatted with [Prettier](https://prettier.io/) 
7. Tests run directly with `jest` (and linted with `eslint`)
8. API requests are proxied through a JavaScript server

### 1. Mastarm & Woonerf
The original motivation for these projects was to consolidate common tooling and libraries across the many JavaScript/React projects. Our focus narrowing removes the need to keep these separate. Additionally, the higher surface area of custom code from using our own tooling to development/build/minify/deploy created bugs over the years that only I could fix and would need to dig deeper into the weeds of Browserify/Babel etc. then I would desire to. 

### 2. Next
Next is a framework for building React websites. The major highlights are it uses Webpack (vs Browserfiy), file system routing (everything in the /pages directory is it's own URL), code splitting, server side rendering, and other optimizations out of the box that should lead to a faster site, with smaller JavaScript builds and make it much simpler to add new pages and code that doesn't bog down the rest of the site.

For example: our Admin page code is loaded with the rest of the site even though it is never used except by administrators, with the new site it will only be loaded when you actually use the admin page.

There are many other changes throughout the site that were necessary because of this change -- like the removal of React Router and how we initialize Redux.

### 3. Now
Next compiles the JavaScript, now deploys and hosts it taking care of the need to manage S3, CloudFront, and the `mastarm deploy` functionality. Additionally all pull requests will automatically be deployed.

### 4. Auth
Redirecting to the login page for verification on first page load was a lingering problem. Auth0 offers a hosted version at a dedicated website (auth.conveyal.com) that we can redirect to when needed. We also are using the prescribed authentication methods so that verifying will only take place when necessary.

### 5. Flow
Types are a "good idea". But once we declared flow bankruptcy due to a the amount of opaque errors we were getting in newer versions of flow it made me realize that we were not using Types in the best way possible. I plan to explore adding TypeScript in the future. But when we do I want to do it more intentionally and take a more holistic approach.

### 6. Prettier
Stripping all flow types left obvious gaps in the code from where they used to be. I had wanted to use an auto-formatter on the code base for awhile and this seemed like the appropriate time.

### 7. Jest / ESLint
Testing / Linting was done via Mastarm. Now it's down directly with the tools we were using before. I was able to speed up both.

### 8. Proxy
Since our API requests now go to a separate server from the web server we would either need to set up CORS settings or proxy the requests through a separate backend. I decided to the go with the latter. Eventually we can do the authentication verification there and route all simple CRUD DB requests there which would greatly simplify the requirements of the Java backend leaving it for only the heavy lifting.


ansoncfit:
New look/feel are good overall.  Starting a laundry list of small issues I notice:

- [ ] URL without query string (e.g. /analysis) returns 500 error page without helpful message/redirect.

Analysis mode:
- [ ] Isochrone colors (especially blue/red overlap) similar to water features
- [ ] Option to clear comparison project/scenario gone
- [ ] Download comparison isochrone button gone
- [ ] Download isochrone button always downloads Baseline

Opportunity datasets:
- [ ] Dot density maps not showing

trevorgerhardt:
@ansoncfit I believe I've addressed all the issues mentioned above except the download button. I had temporarily removed it (due to breaking it with other changes) and didn't add it back for other reasons. 

My thinking goes: do we need the ability to download the comparison isochrone? It may just add clutter for a rarely used feature that can be done in an alternative way -- change the project/scenario, re-fetch, and download. Similarly with the "export comparison analysis" button in regional analyses. 

Removing the buttons does make that specific action "harder". But it also duplicates an action that could be done alternatively and add's an extra button to already cluttered UI panels.

ansoncfit:
Next punch list:
- [ ] Cannot select pattern in convert to frequency modification (`a.setActiveTrips is not a function`)
- [ ] Zeit application error (router) after refreshing a modification detail page
- [ ] Opportunity dataset not clearable on analysis page

PR #885, opened 2019-07-16
v4.9.0

trevorgerhardt:


trevorgerhardt:
Would disabling non-numeric input and inputs over four characters in length do the trick?

abyrd:
Strict format checking works great on dates and times. Thanks @trevorgerhardt.

PR #896, opened 2019-07-26
Add a transit mode selector for add trip pattern modifications

### LINKED ISSUE #806: Set mode (GTFS route_type) for add-trip modifications
I thought there was an existing issue for this, but I couldn't find it.

R5 sets default mode of an add-trip modification to BUS, but in testing, it looks like add-trips modifications are treated as route_type = 0 (TRAM)

trevorgerhardt:
Defaults new Add Trip Modifications to BUS (same as the R5 default). Old modifications will still not have anything set and get the r5 default. Backend change adding a `transitMode` parameter to the AddTripPattern model is necessary for the changes to work (https://github.com/conveyal/analysis-backend/pull/235).

close #806

### DIFF
--- a/lib/components/transit-mode-selector.js
+++ b/lib/components/transit-mode-selector.js
@@ -1,0 +1,32 @@
+import startCase from 'lodash/startCase'
+import React from 'react'
+
+import {ALL_TRANSIT_MODES} from '../constants'
+
+import {Select} from './input'
+
+const toStartCase = o => startCase(o.toLowerCase())
+
+/**
+ * Only show an empty option if no mode has been selected. Do not allow de-selection.
+ */
+export default function TransitModeSelector(p) {
+  const value = parseInt(p.value)
+  return (
+    <Select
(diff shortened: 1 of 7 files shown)


lib/components/transit-mode-selector.js
```diff
@@ -0,0 +1,32 @@
...
+      label='Transit Mode'
+      onChange={e => {
+        const mode = parseInt(e.target.value)
+        if (mode >= 0) p.onChange(mode)
+      }}
+      value={p.value}
+    >
+      {(isNaN(value) || value === -1) && <option value={-1} />}
+      {ALL_TRANSIT_MODES.map((m, i) => (
+        <option key={m} value={i}>
```
ansoncfit: The values are the positions in the `ALL_TRANSIT_MODES` array?  If the order of that array is used, should we add a comment to `constants/index.js` noting that the order implies GTFS route_type values?

PR #900, opened 2019-08-12
Ignore moment locales by default to reduce bundle sizes

trevorgerhardt:
Removes unused Moment locale files which helps reduce gzipped bundle size by 300kb across all chunks.


trevorgerhardt:
@ansoncfit correct, there should be no noticeable changes in the UI. We were not using any of the locale specific features.

PR #904, opened 2019-08-22
Zoom to a modification on click.

### LINKED ISSUE #898: Button to zoom to extents of modification
Allowing an option for this, rather than zooming/panning automatically, fits well with our semi-automated approach

trevorgerhardt:
No longer automatically zoom to modification on navigation. Do it once on click or pattern change.

- [x] Zoom to patterns on pattern change

Screen Shot 2019-08-22 at 6 56 20 PM

fix #898


trevorgerhardt:
@ansoncfit modified the ApplicationDockTitle to use `display: flex` which better handles long strings in the title. This component is used in a variety of places (Analysis title, Modification list title) and so there are a few other components that needed updating.

PR #907, opened 2019-08-27
Update the recommended r5 version

trevorgerhardt:


abyrd:
Note: should only be approved/merged once we've released R5 v4.6.1.

PR #913, opened 2019-09-02
Fix opportunity dataset upload status

### LINKED ISSUE #908: Opportunity dataset upload statuses are not displayed
It appears that the `regionId` is not being passed correctly to the opportunity dataset upload component causing `undefined` to be used when querying status messages. 

This may be causing other unknown behavior.

trevorgerhardt:
Region ids were not being passed correctly when checking for a status, causing a request to be made to an undefined region. This change fixes that error.

## How to test
Upload a new opportunity dataset and verify the status messages are correctly displayed.

fix #908

### DIFF
--- a/lib/__tests__/__snapshots__/reducers.js.snap
+++ b/lib/__tests__/__snapshots__/reducers.js.snap
@@ -65,7 +65,7 @@ Object {
   },
   "queryString": Object {},
   "r5Version": Object {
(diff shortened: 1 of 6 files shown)


abyrd:
I also tested creating a new region (uploading OSM data) which provides less feedback. It just disables the create region button and waits silently for upload/processing to complete. I don't remember if that ever provided more feedback.

PR #915, opened 2019-09-04
Fix bug preventing users from selecting phasing options

### LINKED ISSUE #912: Missing phasing fields in add-trip and adjust-frequency modifications
After "Phase at stop" and "Phase from timetable" are set, users should have the option to select a phase from stop and a phasing offset.  But these aren't appearing in the latest UI, for new or existing modifications.  E.g. https://analysis-staging.conveyal.com/regions/5a2e9df732b98e221fef89eb/projects/5d48529d32b98e762a2a904d/modifications/5d67b94a32b98e443a89bcb8

Also, the first two fields mentioned above used to be clearable, but they aren't any more.

trevorgerhardt:
## Description

The upgrade to the latest version of react-select required new props for capturing the selected value and for setting if the component is clearable. The Phasing select's were not updated correctly during the process. This PR:

1. Makes the Phasing Select boxes clearable
2. Fixes the passed values to be the correct ID's when selected.
3. Disables the auto-clearing of non-existing Timetables and instead shows a warning.
4. Refactors the code to be a bit simpler and use hooks.

fix #912 

## How to test
1. Create an Add Trips modification with two timetables.
2. Phase one timetable to the other timetable and vise-versa.
3. Delete one timetable and ensure that the warning message is now shown for the remaining timetable.

### DIFF
--- a/lib/components/modification/phase.js
+++ b/lib/components/modification/phase.js
@@ -1,104 +1,79 @@
+import {faInfoCircle} from '@fortawesome/free-solid-svg-icons'
 import get from 'lodash/get'
-import React, {PureComponent} from 'react'
+import React from 'react'
 
 import message from 'lib/message'
 import {toString as timetableToString} from 'lib/utils/timetable'
 
+import Icon from '../icon'
 import Select from '../select'
 import {Group as FormGroup} from '../input'
 import MinutesSeconds from '../minutes-seconds'
 
-export default class Phase extends PureComponent {
-  state = {}
+export default function Phase(p) {
+  const {phaseFromTimetable, update} = p
+  const selectedTimetableId = (phaseFromTimetable || '').split(':')[1]
+  const selectedTimetable = p.availableTimetables.find(
+    tt => tt._id === selectedTimetableId
+  )
 
-  static getDerivedStateFromProps(props) {
-    return {selectedTimetable: getSelectedTimetable(props)}
-  }
-
-  componentDidUpdate(prevProps) {
-    if (
-      this.props.phaseFromTimetable !== prevProps.phaseFromTimetable &&
-      !this.state.selectedTimetable
-    ) {
-      this._setPhaseFromTimetable(null) // timetable does not exist
-    }
-  }
-
-  componentDidMount() {
-    if (this.props.phaseFromTimetable && !this.state.selectedTimetable) {
-      this._setPhaseFromTimetable(null)
-    }
-  }
-
-  _setPhaseAtStop = stop => {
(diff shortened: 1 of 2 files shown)


lib/components/modification/phase.js
```diff
@@ -1,123 +1,96 @@
...
+    <>
+      <h4 style={{display: 'flex', justifyContent: 'space-between'}}>
+        <span>Phasing </span>
+        <a
+          href='http://docs.analysis.conveyal.com/en/latest/edit-scenario/phasing.html'
+          rel='noopener noreferrer'
+          target='_blank'
+          title='Learn more about phasing'
+        >
+          <Icon icon={faInfoCircle} />
```
ansoncfit: ℹ️  -- nice.  We should add more links like this.

lib/components/modification/phase.js
```diff
@@ -1,123 +1,96 @@
...
+        </a>
+      </h4>
+      <FormGroup label={message('phasing.atStop')}>
+        <Select
+          isClearable
+          isDisabled={p.disabled}
+          name={message('phasing.atStop')}
+          getOptionLabel={s => s.stop_name}
+          onChange={s => update({phaseAtStop: get(s, 'stop_id')})}
```
ansoncfit: Minor visual issue: once a timetable is selected, all appear selected in the dropdown
image

lib/components/modification/phase.js
```diff
@@ -126,26 +99,22 @@ export default class Phase extends PureComponent {
...
-
-function getSelectedTimetable(opts) {
-  return opts.availableTimetables.find(
-    tt =>
-      opts.phaseFromTimetable &&
-      tt._id === opts.phaseFromTimetable.split(':')[1]
+            ) : (
+              <div className='alert alert-danger'>
+                <strong>Selected timetable no longer exists.</strong> Please
+                clear the timetable or there will be errors during analysis.
```
ansoncfit: If the referenced timetable is deleted, `Phase from timetable` will be cleared already.  Should this instruct the user to clear `Phase from stop`?

image

lib/components/modification/phase.js
trevorgerhardt: That was happening for all of the select boxes. Pushed a fix.

lib/components/modification/phase.js
trevorgerhardt: Updated the text to "Selected timetable no longer exists. Please
                clear or change your phasing options or there will be errors
                during analysis."

Let me know if that works!

PR #916, opened 2019-09-04
Fix r5 version warning

### LINKED ISSUE #914: On region with no regional analyses, message about "completed regional analyses"
During recent testing, on a freshly created region with no regional analyses run, I was seeing the message "You have completed regional analyses with a different routing engine. Using different routing engines may change results." 

To see this, go to this region: https://analysis-staging.conveyal.com/regions/5d6ccad032b98e3060748c0c/regional

You can see it has no regional analyses. Then go to the Analyze tab. The default version is 4.7.0. You should also see the message "WARNING! You have completed regional analyses with a different routing engine. Using different routing engines may change results."

Is this message triggered by regional analyses in the same region, or just in the same organization/group? Is it possible that deleted analyses are affecting the message?

trevorgerhardt:
## Description

When there were no previously used versions of r5 we would display an unnecessary warning saying there was a mismatch. This PR fixes that.

Fixes #914

## How to test
1. Verify no warning message is shown when no previous analyses have been run or the previous analyses have been run with the recommended version (currently v4.7.0)
2. Verify a warning message is shown when previously run analyses have been run with something other than the currently selected version.

### DIFF
--- a/lib/components/analysis/__tests__/__snapshots__/profile-request-editor.js.snap
+++ b/lib/components/analysis/__tests__/__snapshots__/profile-request-editor.js.snap
@@ -70,7 +70,7 @@ exports[`Components > Analysis > Profile Request Editor should render correctly 
     <div
       className="col-xs-12"
     >
-      <Connect(SelectR5Version)
+      <SelectR5Version
         disabled={false}
       />
     </div>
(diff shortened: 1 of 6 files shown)


trevorgerhardt:
We have the minimum version set to v4. Nothing older will be shown as an option.

PR #917, opened 2019-09-10
Multiple route warnings

trevorgerhardt:
## Description

When multiple routes are selected, disallow selecting portions of the route displayed as the modification will be applied to the entirety of each route. This also updates the warning messages shown to reflect that.

Fixes #697 

## How to test
1. Create Adjust Speed modification.
2. Select all routes.
3. Run analysis ensuring the modification was applied for each route in the feed.
4. Repeat for Remove Trips.

PR #921, opened 2019-09-24
Add Union option for aggregation areas

### LINKED ISSUE #862: Option to use features of uploaded shapefile as aggregation areas
When a user uploads an aggregation area, analysis creates a single aggregation area (the union of the uploaded shapefile).  For some use cases (e.g. uploading a shapefile of neighborhoods or city council districts), it would be helpful to have each feature available as a separate aggregation area.

trevorgerhardt:
## Description

Add's two new field's to the aggregation area form. A checkbox to control whether or not the upload will union all of the features and a text field for setting the attribute name when not unioning.

Fixes #862 

Note: also removes an unused variable that ESLint was warning about. This variable was no longer in use from #917 

## How to test
1. Upload a shapefile with multiple attributes, leave the union box checked.
2. Upload a shapefile with multiple attributes, uncheck the union box and ensure that new aggregation areas were created for each attribute.

### DIFF
--- a/lib/components/modification/select-feed-and-routes.js
+++ b/lib/components/modification/select-feed-and-routes.js
@@ -47,10 +47,6 @@ export default function SelectFeedAndRoutes(p) {
   const selectedRoutes = routeIds.map(id =>
     p.selectedFeed.routes.find(r => r.route_id === id)
   )
-  const allRoutesSelected =
-    p.selectedFeed &&
-    selectedRoutes.length > 1 &&
-    selectedRoutes.length === p.selectedFeed.routes.length
   return (
     <>
       <FormGroup>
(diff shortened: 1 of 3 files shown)


trevorgerhardt:
I added the `union` parameter as an additional form field. I realized the backend code checks for a URL parameter instead (similar to the `regionId` parameter). I think this is more appropriate as a form field, so I'll open up a PR to the backend making that small change.

lib/components/modification/select-feed-and-routes.js
```diff
@@ -47,10 +47,6 @@ export default function SelectFeedAndRoutes(p) {
   const selectedRoutes = routeIds.map(id =>
     p.selectedFeed.routes.find(r => r.route_id === id)
   )
-  const allRoutesSelected =
```
ansoncfit: Could you update the PR description with an explanation of this change?

trevorgerhardt: Sure! It was removing an unused variable that wasn't removed from a previous set of changes. ESLint was giving warnings about it.

ansoncfit: Thanks!

PR #922, opened 2019-09-25
Option to process uploaded .csv files into freeform (non-grid) pointsets

### LINKED ISSUE #345: Ability to use non-gridded destinations
For methodological and storage/bandwidth reasons we decided to use gridded destinations, and for maintenance and simplicity we went with only grids, with no freeform point sets. 

There are users / clients who really want travel times to a small / medium number of specific points, not to a huge number of opportunities of a certain class. They often also want travel time matrices to/from the same set of points, or to/from two different sets of points. 

Our back end is totally capable of doing this, it just makes things more complicated. We need to at least decide how we will serve these clients. Perhaps it's command-line support for such jobs, perhaps it's full-blown UI support for them.

Personally I have no problem with having command line tools to enqueue jobs or run them in local non-cluster mode. It enables us to do consulting work with less UI maintenance.

ansoncfit:
## Description

Adds two checkboxes to the opportunity dataset page, allowing singals to the backend that a csv should be converted to a freeform pointset (see https://github.com/conveyal/r5/pull/529):

- [ ] Process as discrete (non-grid) points (beta) [if checked, the checkbox below and an Id field appear]
- [ ] Paired origin/destination file (beta) [if checked, an extra set of lat/lon fields appear]

We expect two common use cases to be:
1. Calculating accessibility or travel times using discrete points as origins (rather than all points in our web mercator grid)
1. Calculating travel times for known origin-destination pairs.

For the former case, it makes sense to specify latitude, longitude, and id columns in the uploaded shapefile.  For the latter case, each row would have a second latitude/longitude (hence the second checkbox)

As a side note, we're getting closer to having OD = origin/destination (instead of the jargony "opportunity dataset")

Fixes #345 (at least for the UI)

## How to test
1. On the opportunity dataset page, enter a name and select a csv file
1. Check the "Process as a discrete..." box
1. Enter lat, lon, and id fields
1. Check the "Paired origin/destination" box
1. Enter secondary lat, lon fields
1. Upload
1. After refreshing, the dropdown should have grid and freeform versions of the file available.

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -279,7 +279,9 @@
     "enterName": "Please enter a new name.",
     "confirmDelete": "Are you sure you would like to delete this dataset?",
     "confirmDeleteSource": "Are you sure you would like to delete all opportunity datasets from this source?",
-    "editName": "Edit Dataset Name"
+    "editName": "Edit Dataset Name",
+    "freeform": "Process as discrete (non-grid) points (beta)",
+    "paired": "Paired origin/destination file (beta)"
   },
   "analysis": {
     "mode": "Mode of travel",
@@ -305,7 +307,7 @@
     "gridFiles": "Select opportunity dataset",
     "latField": "Latitude field",
     "lonField": "Longitude field",
-    "idField": "Id field",
+    "idField": "ID field",
     "travelTime": "Travel time",
     "errorsInProject": "Errors were found in the project:",
     "warningsInProject": "Warnings were found applying the project",
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -305,6 +305,7 @@
     "gridFiles": "Select opportunity dataset",
     "latField": "Latitude field",
     "lonField": "Longitude field",
+    "idField": "Id field",
     "travelTime": "Travel time",
     "errorsInProject": "Errors were found in the project:",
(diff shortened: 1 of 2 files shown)


i18n/en.json
```diff
@@ -305,6 +305,7 @@
     "gridFiles": "Select opportunity dataset",
     "latField": "Latitude field",
     "lonField": "Longitude field",
+    "idField": "Id field",
```
trevorgerhardt: Should this be `ID field`?

lib/modules/opportunity-datasets/components/upload.js
```diff
@@ -63,6 +66,37 @@ export default function UploadOpportunityDataset(p) {
...
 
             <Text name='lonField' label={message('analysis.lonField')} />
+
+            <Checkbox
+              label='Process as discrete (non-grid) points (beta)'
+              checked={freeform}
+              onChange={e => setFreeForm(e.target.checked)}
+            />
+
+            <Checkbox
```
trevorgerhardt: Should this check box be below the `idField` Text box?

trevorgerhardt:
Do we have a section in the docs that we can link to in the form describing these options more thoroughly?

PR #931, opened 2019-10-30
Temporary access group rights for admins

trevorgerhardt:
This PR add's a page allowing administrators to set a temporary access group to help with debugging customer accounts.

It stores the access group in a session cookie which ties the temporary access group to the current browser session. When the browser is closed it will be reset. This allow's for using this feature in incognito mode so it will not affect your default browser session.

Depends on https://github.com/conveyal/analysis-internal/pull/45

## Test plan 1

1. Open a private session in your browser.
1. Sign in as an administrator. 
1. Go to `/admin/set-access-group`.
1. Set the access group to the test group. Ex: `test-group`
1. Navigate to the home page. Ensure that the temporary test group name is displayed.
1. Create/Read/Update/Delete a region, project, and modification.
1. Set the access group back to the admin group.
1. Navigate to the home page. Should see the admin groups resources.

## Test plan 2

1. Open a private session.
1. Sign in as a non-administrator.
1. Ensure that you are unable to go to `/admin/set-access-group`.
1. Confirm inability to navigate to a region for a separate access group.


pages/admin/set-access-group.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  return (
+    <div className='container'>
+      <br />
+      <p>
+        Current access group is: <strong>{accessGroup}</strong>
+      </p>
+      <input ref={inputRef} placeholder='Set access group' />
+      <button onClick={setGroup}>Set group</button>
```
ansoncfit: Could this also redirect to the home page?

trevorgerhardt: Yes, I'll revise the text to say "Set group and redirect to the home page"

trevorgerhardt:
> A few issues I noticed in Chrome:
> 
> Test 1:
> Generally worked, except for a CORS error when trying to view opportunity datasets:
> 
> ```
> Access to fetch at 'https://analysis-staging-grids.s3.eu-west-1.amazonaws.com/5dd2df67d4a1cf0952dab37c/5dd2e243d4a1cf0952dab38c.grid?X-Amz-Security-Token=...' from origin 'https://analysis-staging-git-admin-access.conveyal.now.sh' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
> ```

Does this happen only when using a different access group? 

> Test 2:
> The set access group page (and `/admin` page) can be viewed after signing in with non-admin access group, though setting it doesn't enable access to other resources.
> image

Yes, currently the admin pages are "accessible" but they shouldn't be functional. I'll look into a way to redirect away from them based on the access group along with a warning.

pages/admin/set-access-group.js
trevorgerhardt: Change implemented 👍

trevorgerhardt:
Fixed the CORS error with OD retrieval.

PR #933, opened 2019-11-18
Jsolines

### LINKED ISSUE #863: Remove zero-area artifacts from exported vector isochrones

trevorgerhardt:
Updated Jsolines and all other packages that had updates (using npm-check-updates). In turn, updated snapshots, eslint, and re-formatted with new prettier defaults.

While there are a lot of changes the only functional change should be the removal of 2D holes from the Jsolines update.

closes #863


lib/actions/analysis/index.js
```diff
@@ -209,24 +209,20 @@ export const handleSurface = (error, responses) => {
...
-      setScenarioApplicationWarnings([
-        ...surface.warnings,
-        ...(comparisonSurface ? comparisonSurface.warnings : [])
-      ]),
-      setTravelTimeSurface(surface),
-      setComparisonTravelTimeSurface(comparisonSurface),
-      setIsochroneFetchStatus(false)
-    ]
-  } catch (e) {
-    throw e
```
ansoncfit: How is this error handled now?

lib/components/link.js
```diff
@@ -6,7 +6,7 @@ import {routeTo} from 'lib/router'
 export default function Link({children, to, ...p}) {
   const {as, href, query} = routeTo(to, p)
   return (
-    <NextLink as={as} href={{pathname: href, query}} passHref prefetch>
+    <NextLink as={as} href={{pathname: href, query}} passHref>
```
ansoncfit: What's the effect of this change?

lib/modules/admin/selectors.js
```diff
@@ -6,37 +6,29 @@ export const jobs = state => state.admin.jobs
...
-
-export const allBundlesById = createSelector(
-  workers,
-  workers => {
-    const bundles = {}
-    workers.forEach(w =>
-      (w.bundles || []).forEach(b => {
-        bundles[b._id] = b
-      })
+export const jobsWithWorkers = createSelector(jobs, workers, (jobs, workers) =>
```
ansoncfit: Interesting coincidence that our default accessibility analysis is connecting workers to jobs.

lib/utils/component.js
```diff
@@ -1,12 +1,14 @@
...
 import {createRouter} from 'next/router'
-import {RouterContext} from 'next-server/dist/lib/router-context'
+import {RouterContext} from 'next/dist/next-server/lib/router-context'
 import React from 'react'
 import {Map} from 'react-leaflet'
 import {Provider} from 'react-redux'
 
 import {makeMockStore, mockStores} from './mock-data'
 
+global.Math.random = () => 0.5
```
ansoncfit: What is this used for?

lib/components/link.js
trevorgerhardt: The updated version of `next.js` does not require this anymore. So there should be no effect.

lib/utils/component.js
trevorgerhardt: Next.js creates IDs that are used for page versions and show up in snapshots. I meant to add a comment once that actually worked. Will do.

lib/actions/analysis/index.js
trevorgerhardt: The try/catch was leftover from old code that used the error. Catching and rethrowing immediately is the same as not catching at all. The new version of ESLint picked up on that and showed a warning.

PR #945, opened 2019-12-11
New admin page

trevorgerhardt:
* Add links on the left
* Emphasize jobs and make them expandable
* Don't show all of the workers by default
* Add a simple hook for fetching data with an authentication (instead of using the more complex `fetch-action`)


trevorgerhardt:
> I just tested this on staging with a worker running, and the worker didn't appear in the new admin dashboard. Is that expected?

@ansoncfit they currently show for me! Please try again

PR #950, opened 2019-12-18
Refactoring opportunity dataset docs for clarity

Nate-Wessel:
This is mostly a reorganization of existing content. My goal was to put the more important information first before trailing off into the necessary specifics of implementation and formatting. Hopefully the example CSV better clarifies how these should be formatted, though most of the requirements for that are still given in the text. 

I removed the note about later LODES data as it seems only 2015 data are available at the moment.

There is also some semantic ambiguity here which these changes don't yet address, as that would involve changes to the UI itself. 

As it is, a "**dataset**" seems to be a numeric (count) attribute associated with a spatial distribution, and a "**dataset source**" essentially just is the spatial distribution or spatial dataset which may have several count attributes associated with it. 

I would suggest that the term "*dataset*" in the UI be replaced by **opportunity count** or **opportunity count variable** and "*dataset source*" be replaced by **spatial dataset**. Thus the LODES import is an import of a new **spatial dataset** with numerous **opportunity counts**. 

- **Edit Dataset Name** becomes **Edit Count Variable Name**
- **Delete this dataset** becomes **Delete this count variable** or **Delete this opportunity variable**
- **Delete entire dataset source** becomes **Delete Spatial Dataset** (and associated opportunity vars? ... but that is getting too long for a button) It might be helpful to have another dropdown to allow editing spatial datasets directly without having to do so by way of their associated opportunity count variables.

### DIFF
--- a/docs/prepare-inputs/upload-opportunity-data.md
+++ b/docs/prepare-inputs/upload-opportunity-data.md
@@ -54,10 +54,10 @@ Selecting an existing opportunity dataset from the dropdown menu will give you o
 
 ## LODES dataset import
 
-For projects in the United States, Census data from from the _Longitudinal Employer-Household Dynamics_ ([LEHD](https://lehd.ces.census.gov/)) Program can be imported automatically via the LODES data import function.
+For projects in the United States, employment data from from the _Longitudinal Employer-Household Dynamics_ ([LEHD](https://lehd.ces.census.gov/)) Program can be imported with a one-click import function.
 
 <span class="btn btn-info"><i class="fa fa-group"></i> Fetch LODES (2015)</span>
 
-The [LODES](https://lehd.ces.census.gov/data/#lodes) (_LEHD Origin-Destination Employment Statistics_) dataset provides block-level data on the home and work locations of employed people living in the US. These job and worker counts are made available both as totals and as counts disaggregated by industry, occupation, education level, etc.
+The [LODES](https://lehd.ces.census.gov/data/#lodes) (_LEHD Origin-Destination Employment Statistics_) dataset provides block-level data on the home and work locations of employed people living in the US. These job and worker counts are made available both as totals and as counts disaggregated by industry, earnings, education level, etc.
 
-Note that due to various limitations in the LODES data, 2015 is the latest year made available for import at this time. Data in the 2016 release excludes federal employees and the 2017 release is missing data for Alaska and South Dakota. For more information, see the [LODES technical documentation](https://lehd.ces.census.gov/data/lodes/LODES7/LODESTechDoc7.4.pdf).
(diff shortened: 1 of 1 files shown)


docs/prepare-inputs/upload-opportunity-data.md
```diff
@@ -41,4 +56,4 @@ Selecting an existing opportunity dataset from the dropdown menu will give you o
 
 For projects in the United States, Census block-level data on employment and workforce from [LODES](https://lehd.ces.census.gov/data/#lodes) can be fetched by clicking: <br><span class="btn btn-info"><i class="fa fa-group"></i> Fetch LODES</span>
 
-Note that as of the [August 2019 release](https://lehd.ces.census.gov/data/lodes/LODES7/LODESTechDoc7.4.pdf), 2017 data are missing for Alaska and South Dakota, and 2017 and 2016 data exclude federal employees.
```
ansoncfit: I would like to describe limitations with the current data release. Maybe just add to the existing paragraph "Updated LODES data will be importable from Conveyal Analysis after the Census Bureau has addressed these data issues."

docs/prepare-inputs/upload-opportunity-data.md
```diff
@@ -1,33 +1,48 @@
...
+A [Comma Separated Value](https://en.wikipedia.org/wiki/Comma-separated_values) (CSV) file can be used to represent opportunities as point features. The CSV should have columns for latitude, longitude, and one or more associated opportunity counts. 
 
-Once you have done this, click the upload button to start the upload. Opportunity datasets will be saved with the source name you entered and the attribute/column names of the uploaded files.
+In the following example, a CSV file represents the location of restaurants along with the number of restaurants (1 per point) and the estimated number of employees at each. Such a dataset could be used to represent either opportunities for work or for entertainment. 
 
-After processing is complete, you can refresh the page and see dot-density maps of your datasets, converted to the [analysis grid](../analysis/methodology.html#spatial-resolution) used in Conveyal Analysis.
+```csv
+lat,lon,restaurant_count,est_employees,name
```
ansoncfit: This is a helpful example and it looks great in the docs!

image

One minor suggestion.  I might swap the order of "for work or for entertainment" (to match the count/number of employees order)

docs/prepare-inputs/upload-opportunity-data.md
Nate-Wessel: I've rewritten this section to make it clearer what the LODES data is and where it's coming from. I think I FINALLY understand what the difference is between LEHD and LODES data... I'd often heard the two used as almost interchangeable. Give it a read and see if this section still makes sense. 

That sphinx tool is really pretty handy!

docs/prepare-inputs/upload-opportunity-data.md
Nate-Wessel: Switched.

Nate-Wessel:
Looks good! I don't think I have the permissions necessary to merge this myself, so merge away if you think it's good to go.

PR #952, opened 2019-12-20
Stub out cypress and write sample test

trevorgerhardt:
Add cypress and testing-library/cypress as dependencies and stub out the default folder structure for cypress.


Nate-Wessel:
@trevorgerhardt, is your aim to build out all the end to end tests in this repository rather than a separate one as before?

trevorgerhardt:
> @trevorgerhardt, is your aim to build out all the end to end tests in this repository rather than a separate one as before?

Yes, that would be ideal. Then we can automate the running of them with each pull request. We would still need to maintain the ability to run locally.

cypress.env.json.tmp
```diff
@@ -0,0 +1,4 @@
+{
```
Nate-Wessel: Might want to change this to indicate what fields need to be included in `cypress.env.json`: 
password, authZeroClientId, authZeroDomain.

PR #958, opened 2019-12-31
Scenario names in modifications

### LINKED ISSUE #948: Scenarios should use names as well as numbers when editing modifications
When editing a modification, one is able to choose between scenarios in which the modification will be active. These are visibly identified only by a number, though they also have a name which can be shown by hovering over the checkbox, i.e. the name is already stored in the `title` attribute. 

The name is much more important than the number, which is generated automatically and may be hard to remember. Thus, I think it would be better for the scenario checkboxes to be stacked vertically with the name and number printed to the right. e.g.
* [x] 1. First Scenario Name
* [ ] 2. Second Scenario Name

trevorgerhardt:
## Description

Show the scenario names next to the checkboxes to help users determine which scenario the modifications are active for.

Fixes #948 

## How to test

1. Go to a modification
2. Scroll to the bottom. Check and uncheck scenario boxes.
3. Add a new scenario. Ensure that the new scenario shows up in the modification.


Nate-Wessel:
We may also want to use a normal weight for the text. The bold makes the scenario names heavier than their own title/label.

trevorgerhardt:
Thanks for catching that @Nate-Wessel. Fixed, and reduced the font weight.

Nate-Wessel:
I'm trying to view the preview deployment to review the changes but it's giving the following error:

Callback URL mismatch.
The provided redirect_uri is not in the list of allowed callback URLs.
Please go to the Application Settings page and make sure you are sending a valid callback url from your application.

trevorgerhardt:
@Nate-Wessel I updated the Auth0 configuration so it should work now. I may have changed it by accident the other day. Not sure how it broke.

Nate-Wessel:
@trevorgerhardt I'm still getting the same error. Trying to follow the link to https://analysis-staging-73gd1wug1.now.sh/

trevorgerhardt:
Ah, use the preview links that end in .conveyal.now.sh

https://github.com/conveyal/analysis-ui/pull/958#issuecomment-569972213

PR #960, opened 2020-01-02
Documentation review

Nate-Wessel:
This PR is for some ongoing work on the documentation that I'm doing in parallel with writing tests. Basically as I try to understand each feature in order to define it's expected behaviour, it becomes clear where the documentation is lacking. 

One of the bigger issues I saw with the documentation initially is that it was structured sort of like a learn-by-doing walk-through of the software. There's a place for that, but it meant that some of the structural questions were not getting addressed. E.g. I know how to create and use a project but not necessarily what it is in relation to other components. 

Ultimately, I'd like to see a walk-through type introduction (or two) as well as a top-down specification of components and their various functionality. A reference and a narrative guide.


Nate-Wessel:
I'll try to give an overview of the changes made so far: 

* Mostly, I've just tried to make sure the text reads clearly and easily. Some of this is breaking up heavy sentences, some is moving digressions to less prominent positions, or just being sure to define terms as they appear. 

* I've extended the glossary and made it easier to reference terms using a semantic markup recognized by Sphinx. ``:term:`gtfs` `` rather than `[gtfs](../glossary.md#etc)`. I've tried to be consistent about linking terms where they're first used in a page, usually in the first paragraph. It may look like a lot of links in some places, but IMO this is a stylesheet issue - the actual links are helpful. It wouldn't be a bad idea to de-emphasise these in the stylesheet to distinguish them from standard links. 

* I cleaned up some awkward spacing, trying to keep text together in paragraphs for easier reading. In a few places, I've done this by moving icons inline. 
Screenshot_2020-01-19_10-39-44
In some other places I got rid of button blocks, especially if the button had a really intuitive name already that I could just reference in the text, italicised. 

* I tried to update terminology and keep it consistent throughout. For example, the *add trip patterns* modification was referred to a few different ways before: *add trips*, *add new trip pattern*, etc. Or *GTFS feed* vs. *GTFS* vs. just *feed*.

* I added some new screenshots, or updated them where the interface had changed significantly. I also added captions to many figures where they were missing.

* I reorganized and renamed some of the headings or menu names. For example, **Usage suggestions and details** became **Managing larger projects** as most of the suggestions seemed to be about dealing with long lists of modifications, multiple users, etc. 
I broke a few sections out into their own page where it made sense. I've checked that there shouldn't be any broken links. Changes to headings will mean that some people with IDs in their URLs will land at the top of the page rather than in the correct section.

In general, I've been working from front to back, so There are a lot of changes in the first half of the documentation, trailing off near *managing larger projects* and *regional analysis*. When reviewing this, I would suggest just building the docs for this branch and comparing them to the existing docs side by side. Trying to track every little change might be quite a bit of work.

docs/prepare-inputs/index.md
```diff
@@ -1,65 +1,51 @@
...
+```shell
 osmosis --read-pbf input.osm.pbf \
   --bounding-box left=-79.63 bottom=43.61 right=-79.12 top=43.83 \
   --write-pbf cropped.osm.pbf
 ```
 
-**OsmConvert:**
-
-```
+**OsmConvert:**
```
ansoncfit: Bring the paragraph starting "When creating a region..." up here?

docs/prepare-inputs/index.md
```diff
@@ -71,21 +57,44 @@ When creating a region, the panel will show an osmconvert command pre-filled wit
 
 Note that files larger than 500MB may be rejected on upload. Please contact us if you genuinely need to upload a file of this size, or need assistance in cropping and filtering OSM data.
```
ansoncfit: "Contact us" should be "contact your support team" throughout.

docs/prepare-inputs/index.md
```diff
@@ -71,21 +57,44 @@ When creating a region, the panel will show an osmconvert command pre-filled wit
...
+
 ## Uploading a GTFS bundle
 
-On the main page for a region, upload your first bundle of GTFS feeds by clicking:
-<br><span class="btn btn-success"><i class="fa fa-database"></i> Upload a new GTFS Bundle</span>
+Start by gathering :term:`GTFS` feeds for the transit agencies whose service will be included in your region. A _GTFS feed_ is a set of CSV files inside a `.zip` archive, and a _GTFS bundle_ is a set of one or more GTFS feeds. 
 
-You can also click this icon on the sidebar to access GTFS Bundles:
-<br><span class="ui-icon"><i class="fa fa-database"></i> GTFS Bundles</span>
+If you just created a new region, you will be prompted to upload a GTFS bundle. Otherwise, you can click the database icon (<i class="fa fa-database"></i>) on the sidebar to access your GTFS Bundles at any time.
```
ansoncfit: Nice, helpful rewording.

docs/prepare-inputs/index.md
```diff
@@ -71,21 +57,44 @@ When creating a region, the panel will show an osmconvert command pre-filled wit
...
 
 ## Creating a Project
 
-Uploading and processing a bundle may take several minutes. Once processing is complete, on the Projects page, click:
-<br><span class="btn btn-success"><i class="fa fa-plus"></i> Create new Project</span>
+Uploading and processing a GTFS bundle may take several minutes. Once processing is complete, you should be able to create a new _project_ based on the bundle you uploaded. If you aren't on the projects page already, click the project icon (<i class="fa fa-cubes"></i>) and then _Create new Project_. 
+
+A project is essentially a wrapper around a bundle which associates it with any scenarios and modifications you may create later on. The purpose of projects is to allow multiple users to work simultaneously on the same baseline network without stepping on each other's toes. Later on we'll cover how modifications can be shared between projects which are based on the same bundle. Once created, the bundle associated with a project cannot be changed.
```
ansoncfit: Should it be "others' toes" (plural)?

And "...that are based on the same..."

docs/prepare-inputs/index.md
```diff
@@ -71,21 +57,44 @@ When creating a region, the panel will show an osmconvert command pre-filled wit
...
 
-Again, note that files larger than 500MB may be rejected on upload. The largest GTFS in regular use are below 400MB so a larger file may indicate a problem. Please contact us if you genuinely need to upload a file of this size.
+Again, note that files larger than 500MB may be rejected on upload. The largest GTFS feeds in regular use are below 400MB and most are much smaller than this. A larger file may indicate a problem. Please contact us if you genuinely need to upload a larger file.
+
+If there is no GTFS available for your region, as a workaround you can use a valid GTFS feed from somewhere else in the world, follow the steps below, then attempt to [import route alignment shapefiles](../edit-scenario/index.html#importing-modifications-from-shapefiles) representing service in your region.
 
 ## Creating a Project
 
-Uploading and processing a bundle may take several minutes. Once processing is complete, on the Projects page, click:
-<br><span class="btn btn-success"><i class="fa fa-plus"></i> Create new Project</span>
```
ansoncfit: Again, my preference is to keep the button, especially if it's easy to work in at the end of a paragraph.

docs/prepare-inputs/index.md
```diff
@@ -71,21 +57,44 @@ When creating a region, the panel will show an osmconvert command pre-filled wit
...
+```shell
+osmium tags-filter input.osm.pbf \
+  w/highway w/public_transport=platform w/railway=platform w/park_ride r/type=restriction \
+  -o filtered.osm.pbf
+```
+
 ## Uploading a GTFS bundle
 
-On the main page for a region, upload your first bundle of GTFS feeds by clicking:
-<br><span class="btn btn-success"><i class="fa fa-database"></i> Upload a new GTFS Bundle</span>
```
ansoncfit: My preference is to keep the green buttons -- they make it easier to quickly scan and find an action.

docs/edit-scenario/index.md
```diff
@@ -1,101 +1,32 @@
...
+After selecting a :term:`project`, you will arrive at the modifications page, shown below. Each project has an expandable list of numbered :term:`scenarios<scenario>`  followed by a list of :term:`modifications<modification>` grouped by type. Each modification represents a single operation on the :term:`baseline network` (for example adding a line, or adjusting the speed of an existing line) and can be activated in multiple scenarios. Scenarios are a way of grouping isolated modifications into meaningful units representing changes to the broader network. For example, you could create a scenario representing each of several funding scenarios with one for service cuts, one for extended service and so on.
```
ansoncfit: "units" doesn't feel like the right word to me. Maybe "packages?"

docs/edit-scenario/index.md
```diff
@@ -1,101 +1,32 @@
...
 
 <figure>
   <img src="../img/create-scenario.png" />
-  <figcaption>Initial view in editing mode</figcaption>
+  <figcaption>Initial view in editing mode for an established project</figcaption>
 </figure>
 
-Each modification represents a single operation on the baseline [transport network](../glossary.html#transport-network) (for example adding a line, or adjusting the speed of an existing line) and can be [activated](#activating-modifications-in-scenarios) in multiple scenarios. To add a modification, click
-<br><span class="btn btn-success"><i class="fa fa-plus"></i> Create a modification</span>
+To prevent data conflicts, the system does not support multiple users or browser tabs editing the same project simultaneously. If you expect multiple people in your organization to be using Conveyal Analysis at the same time, you may want to create projects for each of them. Projects based on the same :term:`GTFS bundle` can share modifications by exporting them from one project and then importing into them another (See :ref:`export_mods`).
```
ansoncfit: Last sentence could be "Modifications from one project can be imported into any other project based on the same GTFS bundle." There's no need to do an export step first.

docs/edit-scenario/index.md
```diff
@@ -1,101 +1,32 @@
...
-1. On the map, click to add stops for the new route.  More details on editing route alignments are [here](modifications.html#add-trips).
-1. In the modifications details panel, click: <br><span class="btn btn-success"><i class="fa fa-plus"></i> Add timetable</span>
-1. Optionally, edit the default timetable parameters (e.g. set 5-minute headways for weekdays between 7 and 9 AM) and add additional timetables
-1. To save your changes and return to the main list of Modifications, click this icon at the top of the modification details panel: <br><span class="ui-icon"><i class="fa fa-chevron-left"></i>Modifications</span>
-1. Add more modifications or proceed to [analyze](../analysis/) your scenario.
+This will open a window prompting you to select the [modification type](modifications.html) and enter a name for the modification. Choose a descriptive name - you will generally want to include at least the name of the route created or effected by the modification. After confirming these details, you will be taken to a page displaying options that vary by modification type.
```
ansoncfit: "...route created or affected..." right?

docs/edit-scenario/modifications.md
```diff
@@ -1,64 +1,115 @@
...
+* `Add trip pattern`_
+* `Adjust Dwell Time`_
+* `Adjust Speed`_
+* `Convert to Frequency`_
+* `Remove Stops`_
+* `Remove Trips`_
+* `Reroute`_
+* `Custom`_ (not for general use)
+
+Across all modification types however the basic actions are consistent.
```
ansoncfit: The however feels out of place to me. Just "Across all modification types, the basic actions are consistent:"

docs/edit-scenario/modifications.md
```diff
@@ -1,64 +1,115 @@
...
+<span class="btn btn-warning"><i class="fa fa-stop-circle"></i> Stop editing</span>
 
-To insert a stop into the middle of an alignment, click any part of the alignment. Once created, any stop can be dragged to move it. Clicking on a stop gives you the option to delete it, or convert it to a _control point_ through which the route will pass without stopping. Similarly, control points can be converted back to stops or deleted by clicking on them. It's important to get the alignment approximately correct so that the length of each segment is correct when used to calculate the travel times between stops.
+New route alignments are defined by by an ordered set of stops and control points. The actual route taken between these can either follow existing streets or it can take a straight line between points. It is also possible to combine the two options in one route as when a bus runs on the street but then diverts into a planned busway which is not yet part of the street network. The distance between stops is used to estimate segment travel time on the new alignment so it can be important to as be as accurate with the alignment as possible.
```
ansoncfit: Just "... alignment so it is important to be as accurate as possible with the alignment?"

docs/edit-scenario/modifications.md
```diff
@@ -1,64 +1,115 @@
...
+
+- **Bidirectional**: If this option is checked (the default value), vehicles will travel in both directions along the described geometry. If it is toggled off, vehicles will only travel in the direction the line is drawn. Directed alignments can be useful when there are [couplets](https://en.wikipedia.org/wiki/One-way_pair) or other aspects of the route that don't follow the same alignment in both directions. Directed alignments are also necessary for [phasing](phasing.html). 
+
+- **Follow streets**: This makes the route follow streets. The default is to connect stops or control points with a straight line. This only applies to segments that are actively being edited and will not cause already drawn segments to follow the streets, allowing you to draw part of a route on street and part off-street. The length of each segment is used to estimate travel times between stops, so this feature is very useful for aproximating the length of bus/streetcar routes.
```
ansoncfit: First two sentences could be confusing. I'd suggest "If this option is unchecked (the default value), stops and control points will be connected by straight lines. If it is checked, the route will follow streets. Toggling this setting only applies..."

docs/edit-scenario/modifications.md
```diff
@@ -88,42 +129,41 @@ When removing the beginning of a route, the dwell times at each stop are removed
...
-route/pattern combination will be removed, and the route will be shown in red on the map. Note that the "active in variants" selector always specifies whether the modification is active,
-so in this case implies that the trips will be removed from the selected variants.
+Another common modification is to remove trips. The most common use is to remove entire routes, but it is also possible to remove only specific patterns on a particular route.
+In order to remove trips, create a new _remove trips modification_, and select a GTFS feed, route, and optionally the trip pattern of the trips to be removed. All trips on this route/pattern combination will be removed and the route will be shown in red on the map. Note that the "active in variants" selector always specifies whether the modification is active. In this case it implies that the trips will be removed from the selected variants.
 
 <figure>
-<img src="../img/remove-trips.png" alt="Remove trips" />
+  <img src="../img/remove-trips.png" alt="Remove trips" />
```
ansoncfit: Could you update this screenshot if you get the chance?

docs/edit-scenario/usage.md
```diff
@@ -0,0 +1,83 @@
+# Managing larger projects
```
ansoncfit: Many of the sections on this page are about general usage (hence "Usage suggestions and details") -- not sure the "Managing larger projects" title applies to all of the topics below.

docs/analysis/regional.md
```diff
@@ -7,9 +7,22 @@ Upon selecting a completed regional analysis, you will see a screen like the fol
...
 
-Using the download button, you can save regional analysis results in a raster format (GEOTIFF). These saved files can be opened in GIS to conduct additional analyses or create custom maps. Downloading results also allows you to see the raw [grid cells](methodology.html#spatial-resolution) used for analysis, rather than the smoother interpolated results shown in your browser.
+## Starting a regional analysis
+
+The analysis interface also allows starting a *regional analysis*, which involves repeating an accessibility calculation for every location in a regular grid (See :ref:`spatial_grid`). To start a regional analysis, first set the appropriate parameters using the controls in this view, and confirm that the isochrones and accessibility plots are as expected.
```
ansoncfit: Now that regional analysis has its own page, a couple quick changes may be in order:
- Specify that the "single point analysis interface allows starting..." (now that this is on a different page)
- Move the "Starting a regional analysis" section to the top, above viewing completed results?

docs/analysis/index.md
```diff
@@ -1,41 +1,44 @@
...
-<br><span class="btn btn-info"><i class="fa fa-refresh"></i> Fetch results</span>
+To retrieve results for the origin marker shown on the map, you can either move the marker to a new location or click the *fetch results* button at the top. This will initialize a compute cluster which may take a minute to start up. 
+If this is your first time performing an analysis with a given :term:`bundle<gtfs bundle>`, it may take some time to build the network. This only needs to be done once for each bundle. 
+For more information, see :ref:`why_the_wait`
 
-Once a compute cluster has initialized (which may take [several minutes](../learn-more/faq.html#when-starting-an-analysis-why-does-the-initializing-cluster-message-persist-for-so-long) the first time you use a GTFS bundle), several components will be visible in the analysis mode.
+Once the compute cluster is initialized, you should see an isochrone displayed in blue around your point on the map. If you have selected an opportunity dataset, you will also see a chart showing cumulative accessibility results at selected time and percentile thresholds. You may also select a comparison project and scenario, which will be shown in red.
```
ansoncfit: Should be a :ref: in this section to the new "Options and configuration" page

docs/prepare-inputs/index.md
```diff
@@ -1,65 +1,51 @@
-# Preparing baseline network
+# Preparing a baseline transport network
 
-Start by gathering [GTFS](../glossary.html#GTFS) files for the transit agencies whose service will be included in your scenarios. If there is no GTFS available for your region, as a workaround you can use a valid GTFS feed from somewhere else in the world, follow the steps below, then attempt to [import route alignment shapefiles](../edit-scenario/index.html#importing-modifications-from-shapefiles) representing service in your region.
+Accessibility calculations will be based on the :term:`transport network` that is set up for your :term:`region`. This includes scheduled transit services as well as the region's network of streets, sidewalks, bikeways, etc. Initially you will set up a :term:`baseline network` describing your region's normal transit services. Later on you will likely want to compare alternative :term:`scenarios<scenario>` by creating :term:`modifications<modification>` to the baseline network.
```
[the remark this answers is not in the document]
Nate-Wessel: Changing to: "Initially you will set up a baseline network representing the transit services described in the GTFS feed for your region."

docs/prepare-inputs/index.md
Nate-Wessel: I moved it up. 
Also though, it occurs to me that if the command suggested by the UI used either of the other tools, we could suggest cropping and filtering all in one step. Osmosis would probably be best if Windows doesn't support osmium-tool.

docs/prepare-inputs/index.md
Nate-Wessel: I found a couple more instances of this and changed them as well.

docs/edit-scenario/modifications.md
Nate-Wessel: Updated:
remove-trips

docs/edit-scenario/usage.md
Nate-Wessel: Fair, but I don't think "Usage.." quite fits either. How about **Managing modifications**? Everything does have to do with manipulating/displaying sets of modifications.

docs/analysis/regional.md
Nate-Wessel: Perfectly sensible suggestions. I've incorporated both!

docs/analysis/index.md
Nate-Wessel: Included.

docs/prepare-inputs/index.md
Nate-Wessel: I don't think _each_ can precede a plural. It's an iterator. 

I just included a reference here to the section on _managing modifications_.

docs/prepare-inputs/index.md
Nate-Wessel: Button is resurrected.

docs/prepare-inputs/index.md
Nate-Wessel: Resurrected. 
If we're going to have a lot of buttons, I think we should be more consistent about where they're used.

docs/edit-scenario/index.md
Nate-Wessel: That feels a bit close to "GTFS package" though.

docs/edit-scenario/index.md
Nate-Wessel: Noted. I added a link to the relevant page.

docs/edit-scenario/index.md
Nate-Wessel: I'm not sure what you're getting at.

docs/edit-scenario/modifications.md
Nate-Wessel: Changed.

docs/edit-scenario/modifications.md
Nate-Wessel: I'll change it. 
I had the "can be" in there because distance is just one factor among segment speeds, dwell times, exact departure times, phasing etc. any of which may have much larger accuracy problems.

docs/prepare-inputs/index.md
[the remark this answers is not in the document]
ansoncfit: 👍 maybe feeds plural

docs/prepare-inputs/index.md
ansoncfit: I'm indifferent on whether filtering and cropping are presented as one step or two.

docs/prepare-inputs/index.md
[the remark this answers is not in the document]
Nate-Wessel: Looking at this again, "GTFS bundle" would be the most accurate.

docs/edit-scenario/index.md
ansoncfit: **a**ffected in place of **e**ffected

docs/edit-scenario/index.md
ansoncfit: Is GTFS package a technical term that we or others use?

Nate-Wessel: Hmm. I always say "GTFS package". I'm just noticing now that y'all consistently say "feed" instead. I've definitely heard it both ways. 

It might be OK to use "package" here if we consistently avoid that elsewhere.

docs/edit-scenario/index.md
Nate-Wessel: Ah! I've looked up this distinction a dozen times and still never seem to remember it. Corrected.

docs/edit-scenario/index.md
Nate-Wessel: How about: "A scenario is used for packaging modifications into meaningful groups representing broader changes to the network."
The verb form of "package" shouldn't confuse anything with GTFS feeds.

ansoncfit: Sounds good to me.

abyrd: The term "GTFS package" makes sense since it's an archive of CSV files, but I have never heard anyone say it before. I know "feed" doesn't make sense at all for static data since it's not a stream of individual messages. I almost always hear "GTFS feed", used ambiguously for either an individual file or a series of ZIP files appearing over time for the same operators. I always assumed it was an artifact of RSS feeds being the new hotness around the time GTFS was formulated.

PR #962, opened 2020-01-03
Setup Cypress and Github Actions for adding new tests later

trevorgerhardt:
This PR is mainly more configuration for running Cypress tests both locally and with Github Actions. It also re-enables us to run a production build locally (without `now dev`) using `yarn build && yarn start`. As long as you have `.env.build` set up with local parameters.

This mainly touches test files so I don't think testing on any deployed build is necessary.


cypress/integration/region_spec.js
```diff
@@ -0,0 +1,45 @@
+describe('Set up a new region', function() {
```
Nate-Wessel: These preliminary tests will mostly fail, and may accumulate screenshots somewhere if they are run automatically. It might be best to just comment them out until they can be developed further.

trevorgerhardt: I made some modifications to get these to pass. See here: https://github.com/conveyal/analysis-ui/pull/962/checks?check_run_id=412205424#step:5:112

PR #981, opened 2020-01-16
Allow changing selected phase from timetable without crashing

### LINKED ISSUE #961: Blank screen after clearing a frequency entry pattern when it is referenced by phaseFromTimetable in another entry
**Describe the bug**
A user created an adjust-frequency modification with multiple frequency entries (call them 1 and 2).  Entry 2 refers to Entry 1 via `phaseFromTimetable`.  The user then wants to change the Entry 1's `sourceTrip` and clears it, which causes a blank screen and the console error "Cannot read property 'stops' of undefined"

**Expected behavior**
When phasing references break, we should display a clearer error rather than a blank page.

**Screenshots**
https://app.logrocket.com/7kj4kr/conveyal-analysis-production/s/3-e786d207-bf24-4e6c-9f0a-058f6cc6453c/1?fromTab=latest&t=1577995200378.3787

trevorgerhardt:
Check to see if the timetable exists before accessing the properties. This prevents a white screen
of death.

fix #961

## How to test

See #961

### DIFF
--- a/lib/components/modification/__tests__/__snapshots__/add-trip-pattern.js.snap
+++ b/lib/components/modification/__tests__/__snapshots__/add-trip-pattern.js.snap
@@ -118,7 +118,6 @@ exports[`Component WithContext(AddTripPattern) snapshot(shallow) 1`] = `
     update={[MockFunction]}
   />
   <Timetables
-    allPhaseFromTimetableStops={Array []}
     bidirectional={false}
     modificationStops={Array []}
     projectTimetables={Array []}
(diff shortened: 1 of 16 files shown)


trevorgerhardt:
> These changes fix the originally reported issue. But a WSOD also occurs if you delete an entire frequency entry if it's referenced by another.

Fixed 👍

ansoncfit:
Thanks for the fix. There's still some strangeness -- when stop menus are dropped down, the whole window becomes scrollable (become some divs expand?):

image

trevorgerhardt:
@ansoncfit  I have other changes to the map issue in the "Breaks" branch. If the Phasing issue is fixed here let's merge and ensure that the map height issue is resolved in either the Breaks branch or a subsequent PR.

PR #985, opened 2020-01-21
New breaks

### LINKED ISSUE #973: Regional Analysis comparison map not centered on Zero, extends beyond region bounds
In the regional analysis comparison map, zero (difference in accessibility between scenarios) seems to be treated just like any other number and gets lumped in with other values in what seems like a quantile binning scheme. The documentation indicates that the legend should be centered on zero though with higher relative values in blue, lower in red. 

One of the side effects of this seems to be that whatever bin contains zero extends across the entire map, and actually also beyond the region bounds as in the screenshots below.

Screenshot_2020-01-13_10-21-10

Screenshot_2020-01-13_10-22-59

Zero should not be binned with other values and should display as fully transparent. 
There should probably also be a region bounding box shown on the map and there definitely shouldn't be any values rendered beyond it.

trevorgerhardt:
Previously, breaks were generated over the whole range of grid values. This caused a few issues

1. Zero to be rendered a color, which included parts of the map not inside the analysis area.
2. Colors to overlap in the negative and positive range. There could be reds correlating to positive numbers and blues correlating to negative numbers. 

This change set ensures that 0s are always transparent. In a comparison, this includes the closest positive and negative bins to 0. Additionally, negative bands will always be in the red color range and positive bands will always be in the green color range.

* Removes `browsochrones` dependency
* Removes `gridualizer` dependency
* Separates dot map rendering completely from choropleth rendering

Fixes #973


Nate-Wessel:
Something weird is happening:
Screenshot_2020-01-22_13-20-25
The actual color bins are looking a lot better though!

trevorgerhardt:
> Something weird is happening:
> Screenshot_2020-01-22_13-20-25
> The actual color bins are looking a lot better though!

That should be fixed @Nate-Wessel

ansoncfit:
Some artifacts showing up due to overshoot in interpolation?

image

(https://analysis-staging-git-new-breaks.conveyal.now.sh/regions/a87eb73e-ca9f-48c2-9874-27b1e677b0e6/regional?analysisId=b8bcd7a7-6b0f-494b-bc0a-71604d62c2d0)

abyrd:
It's notable that some 0-1 transitions are hard-edged instead of being smooth, this does seem like something possibly amiss in the interpolation.

Screen Shot 2020-01-25 at 15 00 51

abyrd:
Something odd is happening, if I compare my scenario to my baseline results, I see only negative changes: 
image

If I reverse the layers, comparing baseline to scenario, I see both negative and positive changes:
image

PR #994, opened 2020-01-28
Include reroute in modification bounds

### LINKED ISSUE #959: "Fit map to modification extents" doesn't include reroute extension
When extending a trip pattern with the reroute modification, the new geometry doesn't seem to be considered when automatically zooming to the modification extents. It zooms to the original trip pattern only. It should include both the original and new geometry in the map extent. 

To recreate: 
Create a reroute modification and extend it far beyond the original trip pattern. 
Click the button at the top to "fit map to modification extents".

trevorgerhardt:
Reroute geometries were excluded from modification bound calculations. This change adds them.

fix #959

## How to test

1. Create a modification of each type, with a selected route or a added geometry
2. Click the "Fit map to extents" button.
3. Ensure the entire modification is now shown on the map.

### DIFF
--- a/lib/selectors/modification-bounds.js
+++ b/lib/selectors/modification-bounds.js
@@ -1,8 +1,5 @@
 import lonlat from '@conveyal/lonlat'
-import get from 'lodash/get'
 import {createSelector} from 'reselect'
-
-import {ADD_TRIP_PATTERN} from 'lib/constants'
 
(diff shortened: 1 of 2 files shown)


Nate-Wessel:
This fixes the problem at least for the particular modification that prompted the issue. I haven't tested it on more than that one though.

PR #996, opened 2020-01-28
Clear and load opportunity dataset statuses appropriately for each region

### LINKED ISSUE #990: Opportunity upload status messages persist through page refresh
When uploading multiple opportunity datasets, some of the status messages ("processing", "done") persist after the page is refreshed. 
Screenshot_2020-01-25_12-00-41

That is, they will disappear after a refresh, but all of them (plus a new one) will reappear again when a new dataset is uploaded. This is true even if I do a hard refresh with shift+control+R.

Clicking the `X` will make them go away permanently and clicking the refresh link in the status itself will sometimes throw an error which is visible for a second before the page refreshes:
Screenshot_2020-01-25_12-21-10

The blue processing message in the screenshot above does not have an `X` to click and will not go away. I also saw some of these messages briefly after switching to a different region and uploading data there.

trevorgerhardt:
Load statuses along with OD data on page load which should ensure only seeing statuses for the current region. Also, wait for status to clear before reloading the page (which should fix the error found in #990).

Fixes #990

## How to test

1. Upload OD dataset
2. When complete, leave the status message uncleared
3. Switch to another region's OD page. Ensure no message is shown.
4. Switch back to original region's OD page. Ensure status message is shown.


Nate-Wessel:
Actually, this may just be a Firefox problem - I'm not getting an error message in Chrome.

trevorgerhardt:
@Nate-Wessel still not sure why this was a problem in FF but I think I was able to fix it.

Nate-Wessel:
I tried to reproduce the error message a few times and was successful once. If this is just a timing issue, this is probably good enough, especially if this is just a Firefox problem. 

Something I just noticed though: though reload doesn't actually seem necessary - I can see the new opportunity datasets just fine without reloading the page, both in the dropdown menu and on the map.

trevorgerhardt:
Now was unable to reproduce in FF 🤔 removed the reload completely.

Nate-Wessel:
Hmm.. I tried this a few more times, and also found instances where the new dataset is _not_ added to the dropdown automatically. 

Maybe we should hold onto the refresh prompt?

I suspect the difference may have something to do with processing time - it seems like smaller files are added to the list, but if it is big enough to show a blue processing message, the the dataset doesn't appear once the processing has completed.

trevorgerhardt:
Ok, yea, this is how we should have been doing it the whole time.

Now with the latest change when a OD goes from processing/uploading to done it gets reloaded automatically.

PR #998, opened 2020-01-28
Remove clear bundle button

### LINKED ISSUE #963: "Clear selected bundle" option creates blank screen and unmodifiable modification
Several (but not all) of my test modifications, which had been working before, are now giving me the following message when I try to edit them.  
Screenshot_2020-01-07_10-49-41

For one of these I was brave enough to click the button and was greeted with a blank screen. That modification now appears to be inaccessible because whenever I try to modify it I get the same blank screen. There seems to be data sent from the server, but it fails to render at some point with the following error given in my browser console:
`TypeError: "e.selectedFeed is undefined"`

I tried to get around this by copying the modifications into a new project based on the same feed, but I get the same error when trying to access this particular modification. My thinking was that this might update the feed info attached to the modification.

**To Reproduce**
I'm not sure how to reproduce the problem.

**Expected behavior**
However one gets to the error message in the above screenshot, it should be clearer what it means to 'clear a bundle'. It probably also shouldn't ever make the bundle inaccessible and thus unable to be edited or deleted.

**Desktop:**

- OS: Ubuntu 18.04
- Browser: Firefox, Chrome

trevorgerhardt:
Clearing a bundle causes undefined behavior.

fix #963


ansoncfit:
Could we make this a DEV feature instead of totally removing?

trevorgerhardt:
Good call 👍

trevorgerhardt:
@ansoncfit now shown in DEV mode 👍

PR #999, opened 2020-01-28
Use `get` for finding the feed id

trevorgerhardt:
If a user attempted to create a modification on a slow connection, there is a chance the feeds did
not load yet. Use `get` to prevent an NPE that was being swallowed here. Reported via LogRocket

Discovered here: https://app.logrocket.com/7kj4kr/conveyal-analysis-production/issue/10031519/event/255188332?from=errors

## How to test
On a slow connection...
1. Go to a region
2. Open a project
3. Before the page can fully load the feeds, create a modification

### DIFF
--- a/lib/components/modification/list.js
+++ b/lib/components/modification/list.js
@@ -99,27 +99,24 @@ export default function ModificationsList(p) {
     )
   }
 
-  const create = React.useCallback(
-    ({name, type}) => {
-      return dispatch(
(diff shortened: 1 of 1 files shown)


trevorgerhardt:
@ansoncfit I was technically unable to reproduce the error also. Trying to reverse engineer the stack trace led me to the solution here.

Must be a different cause tho...

PR #1000, opened 2020-01-29
Validate region bounds

### LINKED ISSUE #944: Saving region allowed with invalid bounding coordinates
It's possible to save the region settings (region settings tab) with invalid bounding coordinates entered. I tried entering a northern bound of 4000 and hit save. The map clipped the bounding box to 90N but the change didn't seem to have saved after a few minutes of waiting. 
I also tried saving various non-numeric coordinates and got either an error report or a blank screen. 

Since it seems like there is already some JavaScript validation on the form fields, I'd suggest avoiding this problem by disabling the save button when the form is in an invalidate state. It's already disabled if no changes have been made to previously saved values. 

As seen on Firefox 71.

trevorgerhardt:
Creates a shared form component for ensuring the region bounds are within a maximum and minimum numeric range.

fixes #944


Nate-Wessel:
I notice also that if I use my scrollwheel over one of the fields it will jump by integer values now. I did this by accident and ended up somewhere halfway around the world. This should probably be disabled.

trevorgerhardt:
Hi @Nate-Wessel, thanks for the comments 👍 

I made some changes that should address these issues. Allowing for easier typing, only showing invalid when actually invalid, and disabling the scroll wheel.

Nate-Wessel:
Are you trying a new form validation library? I'm still seeing a few issues: 

The scrolling interaction is gone, but I can still move up or down by integer values with the up/down arrow keys. I would expect this to move my cursor to the start/end of the field if anything. People probably aren't going to ever want to increment values like this for a decimal number field.

The left/right arrow keys don't move the cursor. I have to use the mouse to put it in the right spot. (which is why I was even trying the up/down keys)

The negative sign can't be typed, which may be a disappointment for those in the western and/or southern hemispheres ;-)

If I slyly insert an extra digit, I can sometimes get the form into an invalid state without triggering validation or updating the map. 
Screenshot_2020-01-30_10-10-16

trevorgerhardt:
Ya, trying a new NumberInput that is supposed to handle all of this gracefully and is doing anything but that.

trevorgerhardt:
@ansoncfit addressed both issues 👍

trevorgerhardt:
@Nate-Wessel thanks for bearing with my whack a mole fixes 🤦‍♂

PR #1001, opened 2020-01-30
Prevent overscroll in full screen mode

### LINKED ISSUE #971: Bottom half of map display sometimes doesn't load / isn't shown
I haven't yet figured out what might be causing this, but it sometimes happens that the bottom half of the map display doesn't load or is covered by a grey element. I've seen the problem in both Chrome and Firefox (screenshots below, respectively) multiple times. 

Screenshot_2020-01-11_11-31-11

Screenshot_2020-01-11_10-09-36

When this happens, the map always stops at about the same position ~3/5 of the way down the window. It doesn't seem to be related to a Y tile boundary, because I can sometimes drag the map and the map will smoothly appear/disappear when it crosses that point on the screen. 

The problem always seems to fix itself after a little interaction. In particular, adjusting the window size e.g. by opening the console to debug. Navigating to a new page generally seems to fix it as well.

trevorgerhardt:
fixes #971

### DIFF
--- a/lib/selectors/modification-bounds.js
+++ b/lib/selectors/modification-bounds.js
@@ -1,3 +1,4 @@
+import get from 'lodash/get'
 import lonlat from '@conveyal/lonlat'
 import {createSelector} from 'reselect'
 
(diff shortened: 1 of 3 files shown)


trevorgerhardt:
Boy this is annoying. react-select does some funky stuff.

PR #1004, opened 2020-02-02
Improvements to the convert to frequency modification docs

### LINKED ISSUE #970: Trips not found for speed adjustment modification
I'm trying to implement a simple speed adjustment on all trip patterns of a route. When I run a single point analysis, I get an error reporting that seemingly none of the trips on that route can be found. 

Screenshot_2020-01-11_10-27-20

I was able to reproduce the errors in both Cincinnati and Portland test regions: 
1. Create an *adjust speed* modification
2. Select a route and leave all trip patterns selected
3. enter speed modification (e.g. speed factor of 2) with no segments selected. 
4. run point analysis. 

This should scale the speed on that route by 2 across all segments.

Cincinnati test region/project showing the error:
https://analysis-staging.conveyal.com/regions/5df14ce26844a22caafc8557/projects/5e19e5e0c1f0b40e3af18856

Nate-Wessel:
I ran into some confusion with the _convert to frequency_ modifications while testing, and thanks to some explanation by @abyrd finally figured out what was supposed to be going on. I've updated the documentation to reflect my current understanding, though this needs to be checked for correctness as well as typos. 

Probably a separate issue, but I'll mention it here: if a trip pattern is not replaced by a timetable, it may be very silently deleted entirely. We should issue a warning if a trip pattern is not replaced, and maybe show it in red on the map just like a _remove trip pattern_ modification. 
I've highlighted this watch-out with a note at the end of the section for now. 

fixes #970


Nate-Wessel:
This modification might better be named something like "Replace timetables". 
The "frequency" thing isn't super relevant except that we don't support timetables not based on frequencies. And "convert" isn't true if the original timetable was frequency-based.

trevorgerhardt:
> This modification might better be named something like "Replace timetables".
> The "frequency" thing isn't super relevant except that we don't support timetables not based on frequencies. And "convert" isn't true if the original timetable was frequency-based.

Thanks for the feedback here @Nate-Wessel. That is a great observation.

docs/edit-scenario/modifications.md
```diff
@@ -107,17 +107,36 @@ This modification does not take into account the possibility of increased freque
...
+  <img src="../img/select-trip.png" alt="Selecting a template trip based on total travel time" />
+  <figcaption>A convert to frequency modification with a replacement timetable open</figcaption>
+</figure>
 
-You then [add details to each frequency entry](timetable.html). Repeat this process until you have created frequency entries for all the service you want to retain, in both directions.
+Typically, you will need to create _at least_ two new timetables, one for each direction of travel. 
 
-You can choose to remove all existing trips on the route (the default) or choose to retain trips outside the time windows in which you specify frequencies, which is useful when you are changing the frequency for only part of the day (e.g. increased weekend frequency) and want to retain the existing scheduled service at other times. This is controlled using the "Retain existing scheduled trips at times without new frequencies specified" checkbox.
+.. note:: 
+    If the default "retain trips" option is used and a trip pattern is not specifically replaced by a new timetable, that trip pattern is effectively removed entirely.
```
ansoncfit: I agree we should add a cautionary note. I think the logic could be explained more clearly:

Once converted to a frequency-based route with this modification, any of a route's patterns not represented by a frequency entry are effectively removed. With "retain trips" set to the default value of false (unchecked), these patterns will be removed at all times of day. With "retain trips" set to true (checked), they will be removed when any frequency entry is active.

Nate-Wessel: Sorry it took me a while to get to this - I read your comment initially as saying you had already made the change. 

I hadn't realized that even the 'retain trips' option has so much potential to drop trips - I had assumed this applied only to timetables on the trip pattern itself.

PR #1012, opened 2020-02-04
Select Regional Analysis Display Cutoffs

trevorgerhardt:
Allow selecting a regional analysis display cutoff and percentile.


ansoncfit:
Also, the cutoff and percentile parameter don't seem to be included when using the export to GIS download button.

Issue #1013, opened 2020-02-04
Long modification names change layout


Nate-Wessel:
Long modification names cause the side panel to expand rather than wrapping or being cut off. The name in the example below is a bit silly to prove a point, but I've run into this to a lesser degree with what I thought were reasonable names. 

![Screenshot_2020-02-04_13-28-29](https://user-images.githubusercontent.com/7830465/73774653-5b20fb80-4752-11ea-8634-7f7cc9c8df67.png)

PR #1018, opened 2020-02-05
Fix bug causing long modification names to change layout 

trevorgerhardt:
see #1013

This also re-enables `disableAndDim`


Nate-Wessel:
This mostly looks good - there's just one little hiccup. Every once in a while, I think when the page auto-saves, the title seems to be removed and then reinserted. This isn't very noticeable on dev, but with a wrapped title everything shifts up and then back down again. Just for half-second or less, but it's a little disconcerting.

PR #1025, opened 2020-02-06
Update changelog

trevorgerhardt:
Please update the change log as you see fit.


ansoncfit:
I updated as I saw fit. Not sure if the links to PRs are overkill -- feel free to remove if you think so.

pages/changelog/index.js
```diff
@@ -1,41 +1,59 @@
...
-              <a href='/'>Back to Analysis</a>
-            </small>
-          </h1>
-          <hr />
-        </div>
-      </div>
+    <Stack spacing={10} p={20}>
+      <Heading size='2xl'>
+        Conveyal Analysis Changelog{' '}
+        <Link color='blue.500' href='/'>
```
ansoncfit: Style "Back to Analysis" as a button, or drop it down in size/position? Otherwise it looks like:

image

PR #1027, opened 2020-02-10
Decrease polling delays

trevorgerhardt:
Perceived speed of single point and regional analyses is drastically reduced due to how we poll. Currently we poll 1 second * the retry count (up to 20 seconds) for single point analyses. For regional we poll every minute. Since we implemented these delays the system has become more robust. This PR changes the single point to delay for one second after the first 10 tries and then increase the delay up to 10 seconds after. It also decreases the regional polling delay to 15 seconds.

## How to test

1. Ensure no errors while running a single point request
2. Ensure no errors while running a regional analysis

### DIFF
--- a/lib/actions/analysis/index.js
+++ b/lib/actions/analysis/index.js
@@ -108,7 +108,7 @@ export const fetchTravelTimeSurface = asGeoTIFF => (dispatch, getState) => {
         } else {
           // Wait `seconds` and try again
           setTimeout(() => {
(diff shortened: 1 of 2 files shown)


trevorgerhardt:
> Seems to be working well, quick scenario updates feel faster now. But one extra doubling operation slips through and `if (seconds < 10) seconds *= 2` tops out at 16. I think we could top out at 10 with no real downside.

Changed it to `seconds += 1` 👍

PR #1031, opened 2020-02-11
Disable bundle deletion while bundle is in use

### LINKED ISSUE #1029: Projects from deleted bundle neither deleted nor usable
If a bundle is deleted without first deleting its associated projects, those projects become inaccessible and give an error any time you try to view/edit/use/delete them. 
If it's desirable to keep these around for some reason, rather than deleting them along with the bundle, I would suggest making them non-selectable / de-emphasised in the projects menu and removing them from the options for analysis.

trevorgerhardt:
Projects with a deleted bundle are neither deleteable nor usable. This disables bundle deletion for
bundles in use.

fix #1029

### Todo:
- [x] Allow deleting projects with previously deleted bundles


Nate-Wessel:
I also get a blank screen when trying to select a project with a missing bundle in the edit modifications tab.

trevorgerhardt:
@Nate-Wessel errors should be fixed 👍 

Cypress tests failing atm because staging is not up

lib/components/edit-project.js
```diff
@@ -48,9 +49,15 @@ export default function EditProject(p) {
...
-        <P>{p.bundleName}</P>
         <P>
-          <em>Bundle cannot be changed once a project is created.</em>
+          Bundle cannot be changed once a project is created.&nbsp;
+          <ALink
+            to='bundleEdit'
+            bundleId={p.project.bundleId}
+            regionId={p.project.regionId}
+          >
+            View bundle info here.
```
ansoncfit: I do sometimes use the bundle name displayed in the project settings (so "View info for bundle {p.bundleName} here" or similar might occasionally save an extra click) -- but I guess you need to remove it to fix the core issue here and allow for deletion?  And in any case, maybe the breadcrumbs we've been discussing or another approach would be better for reminding users which bundle they're working with.

lib/components/edit-bundle.js
```diff
@@ -117,6 +120,12 @@ export default function EditBundle() {
             </div>
           )}
 
+          {disableDelete && (
+            <div className='alert alert-info'>
+              This bundle is used by {p.bundleProjects.length} project(s).
```
ansoncfit: I'd suggest moving this message down, so it's right above the delete button. Would `message(...)` get too messy here?

lib/components/edit-bundle.js
```diff
@@ -117,6 +120,12 @@ export default function EditBundle() {
...
+              Deleting is disabled while it is in use.
```
ansoncfit: Perhaps "Thus bundle cannot be deleted until all projects using it are deleted (currently used by {p.bundleProjects.length})."

lib/components/edit-project.js
trevorgerhardt: Correct. There is a correlated server side issue that causes an error if we attempt to fetch a deleted bundle. So making the request at all here is an issue.

PR #1042, opened 2020-02-25
Ensure regional analysis can be deleted while running

trevorgerhardt:
## Description

Deleting a running analysis silently fails. This fixes that by properly passing the analysis id.

## How to test

1. Create a regional analysis. 
2. Delete it while running. Ensure it is deleted.


lib/reducers/region.js
```diff
@@ -26,33 +26,6 @@ export const reducers = {
...
-   * Mark a regional analysis as deleted locally so that it disappears
-   * immediately. It will be marked as deleted on the server soon.
-   */
-  'delete regional analysis'(state, action) {
-    const id = action.payload
-    return {
-      ...state,
-      regionalAnalyses: state.regionalAnalyses.filter(r => r._id !== id)
-    }
-  },
```
ansoncfit: Most of this functionality was previously moved to lib/reducers/regional-analsyes.js, so it's no longer needed here.

PR #1046, opened 2020-02-26
Analysis status

trevorgerhardt:
Display time remaining and the total number of workers running for a regional analysis. Calculates the time remaining in browser based on the number of tasks completed since the last update. Shows 1-3 lightning bolts depending on how many spot instances are spun up (3 bolts for > 100, 2 bolts > 10, 1 bolt for > 0).

Screen Shot 2020-02-26 at 20 30 54

Screen Shot 2020-02-26 at 21 07 25

Needs https://github.com/conveyal/analysis-internal/pull/63

### Todos
- [x]  Only show a time estimate when there is more than one worker or more than five minutes has passed


ansoncfit:
https://xkcd.com/612/

lib/actions/analysis/regional.js
```diff
@@ -39,6 +41,38 @@ export const load = regionId =>
...
+    job.complete,
+    job.total,
+    job.regionalAnalysis.createdAt
+  )
+
+  return {
+    ...job,
+    statusText:
+      job.complete === job.total
+        ? 'assembling results...'
```
ansoncfit: 👍

trevorgerhardt:
> 1. The status remained "starting cluster..." even when multiple workers were running (including when a regional analysis was started after 9 workers were already running)

I was accidentally checking `activeWorker` instead of `activeWorkers`. Now fixed 👍 

Not sure about the other problem, but if there is a way to reproduce it please open a ticket 👌

ansoncfit:
Statuses look good now. My only other suggestion would be to not show ⚡ and status if a job isn't being worked on yet. In the example below, there aren't actually 9 workers processing the top job (they're busy on the bottom one). I'm fine leaving that for a future refinement though.

image

trevorgerhardt:
Thanks for the feedback

Ok here's the logic for the status text:
0 origins complete: `starting cluster...` (no bolts shown)
1+ origin complete: `calculating time...` (bolts are now shown)
5 minutes passed or > 1 active worker: `x remaining` 
complete === total: `assembling results...`

PR #1062, opened 2020-03-30
Move OSM management into Bundle page

trevorgerhardt:
**NOTE: this change requires a DB migration**

## Description

OSM is now bundle specific, not region specific. Bundle's now have an `osmId` used for the `OSM` cache. Old bundles are migrated by setting their `osmId` to be their `regionId`.

Needs https://github.com/conveyal/analysis-internal/pull/68

### After DB migration
1. Run an analysis with an old bundle.

### Set of tests that can be done without DB migration
1. Create a bundle with 4 combinations: 1) New OSM, New GTFS 2) Reuse OSM, New GTFS 3) New OSM, Reuse GTFS 4) Reuse OSM, Reuse GTFS 
2. Verify all can be used for modifications and analysis.


lib/components/create-bundle.js
```diff
@@ -62,82 +83,153 @@ export default function CreateBundle() {
...
-        >
-          {uploading ? (
-            <>
-              <Icon icon={faSpinner} spin /> {message('common.processing')}
-            </>
-          ) : (
-            <>
-              <Icon icon={faCheck} /> {message('common.create')}
-            </>
+    <InnerDock>
```
ansoncfit, line 107: The bundle icon on the sidebar isn't active when this is open.

lib/components/create-bundle.js
```diff
@@ -62,82 +83,153 @@ export default function CreateBundle() {
...
+              <Input
+                id='osm'
+                name='osm'
+                type='file'
+                onChange={e => setFormData(d => ({...d, osm: e.target.value}))}
+              />
+              <FormHelperText>Must be in PBF format.</FormHelperText>
+            </FormControl>
+          )}
+
```
ansoncfit, line 176: Visually, a divider or some differentiation between the OSM and GTFS sections would be helpful.

i18n/en.json
```diff
@@ -40,19 +40,20 @@
   },
   "bundle": {
     "name": "Bundle Name",
-    "explanation": "Bundles are a collection of one or more GTFS feeds that make up the transit layer of a baseline network.",
+    "explanation": "Bundles are a collection of OpenStreetMap data and one or more GTFS feeds that make up the transit layer of a baseline network.",
```
ansoncfit, line 43: Idea: we could use the term "Network bundles" here, and "Built networks" for bundles built by a specific version of R5.

lib/components/create-bundle.js
trevorgerhardt: This is existing functionality. There's no back button here. So in order to get back to the "Bundles" view you can click the sidebar icon here.

This occurs on the modifications page also...but we have a back button there.

We do not have consistent behavior here.

lib/components/create-bundle.js
trevorgerhardt: Working on spacing

lib/components/create-bundle.js
```diff
@@ -62,82 +83,153 @@ export default function CreateBundle() {
...
+                }
+                value={bundles.find(b => b.osmId === formData.osmId)}
+              />
+            </Box>
+          </FormControl>
+
+          {formData.osmId || (
+            <FormControl isDisabled={uploading}>
+              <FormLabel htmlFor='osm'>
+                or upload new OpenStreetMap data
```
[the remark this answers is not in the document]
trevorgerhardt: It is not standard in Analysis UI but clicking on a label in web forms is usually associated with focusing the associated form input.

I have been trying to be better about that, especially since we're using a new testing library that encourages best practices like that.

trevorgerhardt:
@ansoncfit did the migration and tested with new and old files. Let me know what you think 👌

trevorgerhardt:
> I just tried this out on staging, but realized that https://analysis-staging-git-osm-split.conveyal.now.sh isn't reflecting the latest commits. Is there a Zeit configuration change needed?

Hmmm. I'm seeing the latest version. Maybe a hard refresh is needed?

> I also saw a failure when trying to view a completed regional analysis: `Uncaught ReferenceError: setImmediate is not defined` (but maybe that has been resolved by recent changes?)

Whoops, apparently that's not cross browser compatible. Will remove 👍

lib/components/bundles.js
```diff
@@ -1,16 +1,42 @@
...
   return (
-    <InnerDock className='block'>
-      <legend>
-        <Icon icon={faDatabase} /> GTFS Bundles
-      </legend>
-      {p.children}
+    <InnerDock style={{width: '640px'}}>
+      <Stack spacing={8} p={8}>
+        <Heading size='lg'>
+          <Icon icon={faDatabase} /> GTFS & OSM Bundles
```
ansoncfit, line 20: I expected this to use `message`

lib/components/create-bundle.js
```diff
@@ -62,82 +92,214 @@ export default function CreateBundle() {
...
+              onChange={onChange('name')}
+              placeholder='Name'
+            />
+          </FormControl>
+
+          <Stack spacing={4}>
+            <Heading size='md'>OpenStreetMap</Heading>
+
+            <Text>
+              Reuse existing OSM data from other bundles or upload new data.
```
ansoncfit, line 157: Edit this and the rest of the instructions/headings in this component to use `message()`?

lib/components/create-bundle.js
```diff
@@ -62,82 +92,214 @@ export default function CreateBundle() {
...
+            </Text>
+
+            <Heading size='sm'>
+              Crop with osmconvert
```
ansoncfit, line 161: I'd move the crop commands under "Upload new OpenStreetMap data", and possibly have them collapsed by default together with the upload form. That reduces visual clutter for the common use case (reusing existing OSM). It is also consistent with the ordering implied above "Reuse existing OSM data... or upload new data."

lib/components/create-bundle.js
```diff
@@ -62,82 +92,214 @@ export default function CreateBundle() {
...
+                  {bundles.map(b => (
+                    <option key={b._id} value={b.osmId}>
+                      {b.name}
+                    </option>
+                  ))}
+                </Select>
+              </FormControl>
+            )}
+
+            {formData.osmId || (
```
ansoncfit, line 217: Hiding upload when existing selected 👍

lib/components/create-bundle.js
```diff
@@ -62,82 +92,214 @@ export default function CreateBundle() {
...
+                west: bounds.west
+              })}
+            </Code>
+
+            {formData.osm || (
+              <FormControl isDisabled={uploading}>
+                <FormLabel htmlFor='osmId'>
+                  Use existing OpenStreetMap data from
+                </FormLabel>
+                <Select id='osmId' name='osmId' onChange={onChange('osmId')}>
```
ansoncfit, line 206: I'm used to seeing an ✖ to clear a selected option, and that would be cleaner here and in the forms below if it's not too much effort. Here, selecting the blank row in the form works as an alternative, but I think it could be less intuitive for users.

PR #1063, opened 2020-03-31
Template OSM crop/filter commands

ansoncfit:
This PR cleans up the template OSM prep commands, and adds a template for osmosis, which should save advanced users a few seconds when setting up new regions.

Also, added/replaced links to the user manual with consistent ℹ icons.

I expect this will conflict with some of the osm creation pages.  These template commands should be copied over to the region settings page (and maybe removed from the region creation page?)

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -124,6 +124,8 @@
     "boundsNotice": "Bounds are snapped to points on our regular grid.",
     "updatesDisabled": "OSM updates have been temporarily disabled and will be moved to the bundle management page.",
     "r5Version": "Analysis backend version",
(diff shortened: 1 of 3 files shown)


trevorgerhardt:
Would you mind re-doing these changes off of the branch for this PR? https://github.com/conveyal/analysis-ui/pull/1062 They would go in the `lib/components/create-bundle` component.

ansoncfit:
Superseded by commits to #1062 

@trevorgerhardt I decided to move the command templates to the edit-region component instead, because that's where region bounds would be edited.

PR #1065, opened 2020-04-05
Cypress scratch

trevorgerhardt:
Create a test suite that can be run locally with no prior set up.


Nate-Wessel:
This should be ready to merge. All tests are passing consistently when I run them locally. There's more to come of course, but that can be added to another PR.

PR #1070, opened 2020-04-16
User Manual update for mid-April release

ansoncfit:
Network building revisions and other misc. updates


docs/prepare-inputs/index.md
```diff
@@ -1,27 +1,42 @@
...
-The process of downloading and processing OSM data into an appropriate format is covered in the next section. For now, if all looks good with your region, you should be able to click the _Set up a new region_ button. After a bit of time for uploading and processing the OSM data, you should be prompted to upload a GTFS bundle.
+## Creating a network bundle
+
+Conveyal Analysis requires a :term:`network bundle`, which consists of street and transit data.
+
+To create a network bundle, click on the Network Bundles icon (<i class="fa fa-database"></i>), then click:
+
+<span class="btn btn-success"><i class="fa fa-plus"></i> Create a new network bundle</span>
+
+In the panel to create a new network bundle, first enter a name. We recommend staying organized by basing the name on the time period and/or services included, such as "April 2020 - Transit and Commuter Rail Feeds."
```
trevorgerhardt, line 25: Not related to this PR, but this makes me wonder if we should ask for more meta data during bundle creation. 

We interpret service from the GTFS feeds, but maybe they should be set manually? 
Are there tags or categories that are common that would help with that?

ansoncfit: The Resource changes to display created on and created by metadata will be helpful here.

Nate-Wessel: Same for OSM data, now that that can be managed. It would be nice to be able to add some info aside from the bundle name. e.g. date it was pulled from OSM, any modifications made to it, etc.

ansoncfit: Prior to the network bundle changes, we had an OSM description form field in the UI. Maybe we could bring that back, with suggested text including "Date extracted from OSM, offline modifications made before upload" etc.

PR #1073, opened 2020-04-17
Delete region locally before sending

trevorgerhardt:
- The culprit here was the `checkLoadStatus` function that was left over from when OSM was uploaded and the region creation took longer than one `fetch`.
- Also fixes a bug that caused opportunity dataset's names not to show

### DIFF
--- a/lib/components/edit-region.js
+++ b/lib/components/edit-region.js
@@ -37,8 +37,11 @@ export function EditRegion(p) {
 
   // Delete region action
   function _delete() {
-    dispatch(deleteRegion(region._id))
-    router.push('/')
+    // Felete that region
+    return dispatch(deleteRegion(region._id)).then(() => {
+      // Then go to the region selection page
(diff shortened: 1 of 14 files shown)


Nate-Wessel:
This seems to be working. I'm not able to reproduce the error.

lib/components/edit-region.js
```diff
@@ -36,11 +37,11 @@ export function EditRegion(p) {
...
   // Delete region action
   function _delete() {
-    if (window.confirm(message('region.deleteConfirmation'))) {
-      dispatch(deleteRegion(region._id)).then(() => {
-        router.push('/')
-      })
-    }
+    // Felete that region
+    return dispatch(deleteRegion(region._id)).then(() => {
+      // Then go to the region selection page
```
ansoncfit, line 42: Thanks for the explanatory comments.

PR #1082, opened 2020-04-24
Remove version restriction when for custom r5

trevorgerhardt:
- [ ] Revert loosening restrictions in the UI element
- [ ] Allow editing r5 version in the custom JSON editor to anything the user wants


trevorgerhardt:
Closed in favor of #1095

PR #1090, opened 2020-05-16
Cypress analysis

Nate-Wessel:
This contains incremental changes and additions to the Cypress tests.

Notable changes: 
- Cypress version upgrade
- Load visual regression testing plugin (not used yet though)
- Some assorted steps to reduce test flake
- Very basic single-point analysis test introduced


cypress/support/commands.js
```diff
@@ -31,16 +32,16 @@ function createNewRegion(regionName) {
   cy.fixture('regions/' + regionName + '.json').then((region) => {
     cy.findByLabelText(/North bound/)
       .clear()
-      .type(region.north, {delay: 0})
+      .type(region.north, {delay: 1})
```
trevorgerhardt, line 35: Please add a comment explaining why there's a delay of 1 here.

cypress.json
```diff
@@ -8,5 +8,6 @@
     "password": "fakepassword",
     "authZeroClientId": "",
     "authZeroDomain": ""
-  }
+  },
+  "dataPrefix": "CYP_"
```
trevorgerhardt, line 12: 👍

cypress/integration/local/modifications.js
```diff
@@ -91,26 +95,56 @@ describe('Modifications', () => {
...
+      // add a timetable
+      cy.findByText(/Add new timetable/i).click()
+      cy.findByRole('alert', {name: /needs at least 1 timetable/}).should(
+        'not.exist'
+      )
+      // add to scenario
+      cy.findByLabelText(/Default/).uncheck({force: true})
+      cy.findByLabelText(/scratch scenario/).should('be.checked')
+      //
       cy.deleteThisMod()
```
trevorgerhardt, line 148: Can this be done in an `afterEach` section?

cypress/support/commands.js
Nate-Wessel: The default delay is 10 (milliseconds), so this is still a speed-up. It was working at 1 for a week or two and I tried taking it down to zero, but that seemed to cause some semi-random errors in the region bounds, probably related to autocorrection on the inputs. This change just put it back where it had been a couple changesets ago.

cypress/integration/local/modifications.js
Nate-Wessel: I had been cleaning up this test specification by deleting the whole project (and thus the modifications). When I refactored the test setup to store intermediate object UUIDs, I switched that off temporarily because I wasn't ready to delete the project ID. I'm planning to bring that back in the after hook, but just haven't quite got there yet.

PR #1092, opened 2020-05-22
Latest round of Cypress tests

Nate-Wessel:


cypress.json
```diff
@@ -7,7 +7,9 @@
     "username": "testing@conveyal.com",
     "password": "fakepassword",
     "authZeroClientId": "",
-    "authZeroDomain": ""
-  },
-  "dataPrefix": "CYP_"
+    "authZeroDomain": "",
+    "region": "scratch",
+    "dataPrefix": "CYP_",
+    "freshSetupEachRun": true
```
trevorgerhardt, line 13: Let's rename this `clearTestData`

cypress/integration/local/all.js
```diff
@@ -0,0 +1,7 @@
+import './regions'
+import './bundles'
+import './opportunities'
+import './projects'
+import './scenarios'
+import './modifications'
+import './analysis'
```
trevorgerhardt, line 7: After looking at this again I think this should be avoided. This style is pretty against the standard test runner setup.

package.json
```diff
@@ -130,7 +130,7 @@
     "@babel/preset-env": "^7.9.5",
     "babel-eslint": "^10.1.0",
     "babel-jest": "^25.3.0",
-    "cypress": "4.6.0",
+    "cypress": "4.7.0",
```
trevorgerhardt, line 133: This is causing the build to break. Use the `^` style when referencing versions. `"cypress": "^4.7.0"`

cypress/support/commands.js
```diff
@@ -6,30 +6,124 @@ Cypress.Cookies.defaults({
...
-Cypress.Commands.add('setupRegion', (regionName) => {
-  // set up the named region from fixtures if necessary
-  cy.task('touch', pseudoFixture(regionName))
-  cy.readFile(pseudoFixture(regionName), {log: false}).then((IDs) => {
-    if ('regionId' in IDs) {
-      cy.visit(`/regions/${IDs.regionId}`)
-      cy.contains(/Create new Project|Upload a .* Bundle/i)
-    } else {
-      createNewRegion(regionName)
+before('Optionally wipe configured state', () => {
```
trevorgerhardt, line 16: There should not be a `before` statement here. This should only be for setting up commands.

trevorgerhardt: Repeating the same `before(` in each file that needs should be the way to go.

package.json
trevorgerhardt: It causes this error: https://github.com/conveyal/analysis-ui/pull/1092/checks?check_run_id=728322156#step:5:40

trevorgerhardt:
One more note, please ensure that this passes `yarn test` locally.

cypress.json
Nate-Wessel: Roger that. Might be a bit of a misnomer though since test data is left in place after a test run, only getting cleared before a subsequent run. That's the suggested way of doing things from the Cypress docs. Agree that the name is awkward as is though.

package.json
Nate-Wessel: Aha. I wondered why things were failing earlier than usual. 

Looking at this, I've just noticed that `cypress-file-upload` is in dependencies rather than devDependencies. That's not right, is it?

cypress/support/commands.js
Nate-Wessel: I think it should actually be in `support/index.js` since this leaves the before hook in the "[root suite](https://mochajs.org/#root-level-hooks)" for any test specification file. That way it doesn't need to be repeated. `commands.js` is just imported by `index` anyway.

Nate-Wessel:
> One more note, please ensure that this passes `yarn test` locally.

Some warnings, no errors. 

Regarding the lint warning, (`Test is missing function argument`), I had a few tests for a while with an empty callback, just as a placeholder for tests I was meaning to write. Apparently the proper way to note this in Cypress is to leave out the callback, and then the runner marks the test as pending rather than passing.

cypress/integration/local/all.js
Nate-Wessel: I totally agree, but the alternative seems to be setting up any needed components fresh for each and every test file. The scenario tests for example take ~5 seconds if everything already exists, but 50 if it needs to fully set up the region, bundle, and project first. 

I'm sure there's a faster way to do the setup, but that would probably involve writing some code to seed the database directly rather than going through the UI.

package.json
Nate-Wessel: It looks like something is still going wrong here. I moved `cypress-file-upload` to `devDependencies`, but that wouldn't be causing the current cypress version problem, I don't think. Let me know if I shouldn't have moved it - I assumed I issued the wrong yarn command when I upgraded it, but it looks like it was there before I upgraded.

cypress.json
trevorgerhardt: Ah, I found those docs in Cypress. Let's name it based on the terminology they use: `resetDataBeforeEach`

trevorgerhardt: `resetDataBeforeEachRun`

package.json
trevorgerhardt: Try deleting the `yarn.lock` file and then running yarn again and checking in the new `yarn.lock` file

cypress/support/commands.js
trevorgerhardt: Ah. I see. Found that here: https://docs.cypress.io/guides/references/best-practices.html#State-reset-should-go-before-each-test

cypress/integration/local/all.js
trevorgerhardt: Ok. Leave it for now then. May change in the future.

trevorgerhardt:
@Nate-Wessel once we can figure out the build issue here and tests pass we can merge 👌

package.json
Nate-Wessel: Ok. It builds, but now there are some failing unit tests. 

Looking at the logs, I'm also noticing that cypress seems to be running [all tests](https://github.com/conveyal/analysis-ui/pull/1092/checks?check_run_id=738955320#step:6:188) in `integration/`, not just `integration/local/all.js`. This means we're running all local tests twice and trying to log in to localhost?

Nate-Wessel: I'm pretty sure the jest test failures we're seeing now are due to a bunch of other packages getting upgraded within their specified version ranges when I rebuilt the yarn.lock file, and not to anything related to Cypress. I spent a little time trying to debug things but I think I'm about as likely to cause more problems as to fix them if I go poking around in unfamiliar code.

PR #1096, opened 2020-05-25
Grid documentation + fixing broken formatting

Nate-Wessel:
Adds mention of the .grid opportunity upload/download file format to the documentation.


Nate-Wessel:
Regarding formatting issues, I think what we're using in the docs now is a weird hybrid of markdown and restructured text. We got away with this for a while, but now something is checking the formatting on a pre-commit hook and insisting that `.md` files don't have `.rst` features. That's broken some formatting on pages that have recently been edited. If we're sticking with Sphinx, we should probably move fully to RST. I think the main difference would be that a few things like headings and images would be differently formatted. Otherwise, the two formats are largely compatible. In the meantime, we can and do have a mix of both filetypes.

Nate-Wessel:
I think I've caught all the accidental formatting and referencing errors. 

One thing worth noting here is that I also changed four references from "gtfs bundle" to "network bundle". These were indicated as broken references to a glossary term that was updated. There are probably other (unlinked) references to "GTFS bundle" that should be updated, but I leave that to another PR.

PR #1097, opened 2020-05-31
Analysis Page Redesign

trevorgerhardt:


lib/components/analysis/advanced-settings.tsx
```diff
@@ -0,0 +1,249 @@
...
+          value={profileRequest.workerVersion}
+        />
+
+        <CustomBoundsSelector
+          isDisabled={disabled}
+          profileRequest={profileRequest}
+          regionalAnalyses={regionalAnalyses}
+          regionBounds={regionBounds}
+          setProfileRequest={setProfileRequest}
+        />
```
ansoncfit, line 79: Having the profile request settings on a 3 column grid, and this on a 2 column grid, makes these advanced settings feel more prominent than they should be. Could be resolved by bringing the customize bounds button down from the label level to the main input level?

lib/components/analysis/single-point-settings.tsx
```diff
@@ -0,0 +1,521 @@
...
+      ) : (
+        <Heading size='md' color={`${color}.500`}>
+          Select a {isComparison ? 'comparison ' : ''}project
+        </Heading>
+      )}
+
+      <Stack spacing={1} isInline>
+        <DownloadMenu
+          isComparison={isComparison}
+          isDisabled={!hasResults || settingsHaveChanged}
```
ansoncfit, line 377: Minor issue: I think these weren't disabled the first time I was building a network, so the download buttons could be clicked (causing errors)

lib/components/analysis/advanced-settings.tsx
trevorgerhardt: Hmmm. I really do like the smaller button inline with the label. Do they really feel that prominent at the bottom? Could just make it a two column grid also.

PR #1102, opened 2020-06-10
Cypress modifications refactor and general speed up

trevorgerhardt:
### Inline assets to speed up tests. 

Each running Cypress instance does not hit the cache for CSS and image assets. In production we would retrieve them from globally used caches. By inlining them we ensure they come from the same location our code does and it dramatically speeds up tests and running locally.

### Cypress modifications test refactor

- Reuse a standard naming scheme
- Prefer `const` over `let`
- Remove use of `random` (does not create reproducible tests)
- Do simple CRUD test for each mod type
- Move modification specific commands into the modification file


trevorgerhardt:
> I get an error message on the first test:
> 
> ```
> com.fasterxml.jackson.databind.exc.InvalidTypeIdException: Could not resolve type id 'add-streets' as a subtype of `com.conveyal.analysis.models.Modification`: known type ids = [add-trip-pattern, adjust-dwell-time, adjust-speed, convert-to-frequency, custom, remove-stops, remove-trips, reroute] at [Source: (String)"{"name":"CYP_MOD_Add Streets_simple_1591799569110","projectId":"5ed912fb7bd56172baaed100","type":"add-streets","variants":[true,true,true],"allowedModes":["BICYCLE","CAR","WALK"],"bikeTimeFactor":1,"bikeLts":1,"carSpeedKph":30,"lineStrings":[],"walkTimeFactor":1}"; line: 1, column: 98]
> ```
> 
> I think I need a new backend jar file for testing this locally?

Ah, correct. I will get you one.

PR #1103, opened 2020-06-10
June release fixes

trevorgerhardt:
- Fix bookmark creation
- Save single point settings when you navigate away (after fetching results)
- Standardize heading sizes
- Fix region name sorting
- Fix regional analysis page (needed to update the input hook)


trevorgerhardt:
> If you go to the analysis page, the initial projectId in the profileRequest is null, which leads to an error when fetching results.

Fixed in the latest version 👌 

Since it remembers your settings per region there was only ever a chance that would happen the first time.

PR #1109, opened 2020-06-13
Modify controlled input hook to prevent resetting cursors in inputs

### LINKED ISSUE #965: Time input fields in timetable window too quick to correct format
When adding a schedule to a new trip pattern, the time input fields can be a bit too aggressive at correcting the format of the input. For example, let's say I want to change the default frequency value of `00:10:00` to `00:30:00`. I try to do this by placing my cursor after the `1`, deleting it, and adding `3`. But by the time I type `3`, it's moved my cursor to the end and I wind up with `00:00:03`. 

It will also often overshoot at the end. I.e let's say I try to make the change by deleting characters from the end until I just have `00:`. I'll then enter the remaining `30:00` but wind up with `00:30:000`.

I'm guessing that whatever form validation is going on here probably just needs to be moved from an `onchange` to an `onblur` event to let the user finish typing before making corrections.

trevorgerhardt:
- Formatted timetable inputs for minutes-seconds should no longer reset the cursor
- Fixes Route Color input
- Memoizes a bunch of components (was profiling the application and saw that we could use it in a few places)

fixes #965


Nate-Wessel:
> This pull request is being automatically deployed with Vercel ([learn more](https://zeit.ink/github-learn-more)).
> To see the status of your deployment, click below or on the icon next to each commit.
> 
> mag Inspect: https://vercel.com/conveyal/analysis-staging/g13gwfljz
> white_check_mark Preview: https://analysis-staging-git-timetable-form-fixes.conveyal.vercel.app

These preview links don't work for me any more. Can I safely assume that `https://analysis-staging-git-timetable-form-fixes.conveyal.now.sh` (working previous link format) correctly references the same thing?

Nate-Wessel:
It's worth noting that these time fields seem to always indicate a valid state, i.e. with the blue border. They should check for valid input and highlight problems in red like other form fields, though without correcting/changing things while the input element is focused.

PR #1119, opened 2020-06-19
Run Analysis Backend for automating Cypress Tests

trevorgerhardt:
The integration workflow now sets up a local backend server to run for each pull request.

These are running the local version of the Cypress tests only.


Nate-Wessel:
Looking at the logs, it appears three Cypress tests actually failed. 
https://github.com/conveyal/analysis-ui/pull/1119/checks?check_run_id=788129189#step:9:2812

trevorgerhardt:
> Looking at the logs, it appears three Cypress tests actually failed.
> https://github.com/conveyal/analysis-ui/pull/1119/checks?check_run_id=788129189#step:9:2812

Not entirely sure what's happening there. Seems to be inconsistent behavior.

trevorgerhardt:
@Nate-Wessel all passing now. But I removed part of the region test that compared the map bounds. It wasn't consistently passing, especially during CI.

Nate-Wessel:
> @Nate-Wessel all passing now. But I removed part of the region test that compared the map bounds. It wasn't consistently passing, especially during CI.

That part depends mostly on whatever external API is behind the location search bar. It's always worked on my machine though, so I wonder if the instance running in CI has a working internet connection?

trevorgerhardt:
> That part depends mostly on whatever external API is behind the location search bar. It's always worked on my machine though, so I wonder if the instance running in CI has a working internet connection?

The tests using location search are working now. It's the test that does the `map.getBounds().contains...` that passes/fails inconsistently.

trevorgerhardt:
@Nate-Wessel looks like Cypress solved the issue already: https://github.com/cypress-io/github-action/issues/168

Issue #1104, opened 2020-06-10
Copy modification doesn't always copy latest changes


Nate-Wessel:
If I create some changes to a previously saved add trip modification and then quickly copy the modification before it has a chance to auto-save, the copy ends up being a duplicate of an older state of the modification. I verified this with changes to route geometry and timetables. E.g. create a new timetable in your modification and ten quickly copy it - the copy may be missing a timetable.

It probably just needs to trigger a save right before the copy operation.

Issue #1057, opened 2020-03-04
Give clearer indication when ModificationGroup is collapsed


ansoncfit:
If the number of modifications of a certain type exceeds 4, the modifications of that type will be collapsed in the modification list.  See:

https://github.com/conveyal/analysis-ui/blob/1969c731882d9213d5bc0295f95e2893dedc4e0b/lib/components/modification/list.js#L193

When a single modification group is collapsed at the top of the list, it can be easy to miss. Users have reported not knowing whether an import of route alignment shapefiles was successful; if there are more than 4 imported routes, the application redirects from the import page to the modification list with no indication the operation succeeded (other than the easy-to-miss appearance of a new, collapsed modification group).

This is exacerbated by the current default of hiding modifications on the map (see discussion in #975).

Showing success "toasts" when operations succeed could also address the reported issue (along with #956)

ansoncfit:
Two alternatives:

Always show ~2 modifications per type. Expand/collapse any remaining with a "show more" or "show fewer" button below.  But duplicating that button could be a waste of vertical space.

Or replace the simple expand caret (in modification type groups, and the scenario menu) with "Show >"

Issue #975, opened 2020-01-15
Add ability to show/hide all modifications


Nate-Wessel:
I think this issue is the opposite of #109, where the problem was that there was no way to hide all modifications. Now there is no way to _show_ all modifications. Each has to be clicked individually.

The hide-all icon (below) should be paired with a show-all icon.

![Screenshot_2020-01-15_13-00-08](https://user-images.githubusercontent.com/7830465/72458736-28a96180-3797-11ea-870c-2f70eeae60a0.png)

trevorgerhardt:
We intentionally do not have this. In instances where projects end up with hundreds of modifications, attempting to show them all at once can crash the application.

ansoncfit:
Projects with hundreds of modifications are the exception, not the rule. 

Could we add some conditional logic, so that the visibility icons for the project and each scenario would show all modifications up to some limit x and show a warning that only x are shown if more are requested?

On a related note, with such limits in place, we could safely return to the practice of showing modifications on the map by default (which would also help with #1057). For users who aren't aware that modifications are saved automatically, it's a bit disconcerting to navigate back to the modification list after creating a modification and not see what they just created.

PR #1120, opened 2020-06-22
Modifications list changes

### LINKED ISSUE #956: Copy modification should give feedback when clicked
When editing a modification, clicking the copy icon gives no visual feedback to the user. It's not clear that it's done anything until you save and return to the modifications list. 

It might make sense to prompt the user for a name to give the copy rather than automatically creating `name (copy)`. We could always autofill the dialogue with the default ` (copy)` name so they can just click through if they want.

trevorgerhardt:
This change refactors the modification list display to be more compact and introduces an `Editable` component for allowing inline editing of text fields like names and descriptions to prevent needing to show/hide an entire input.

closes #956, #975, #1104 and #1057


lib/components/select-project.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+          <Button
+            isFullWidth
+            onClick={routeToBundleCreate}
+            variantColor='green'
+          >
+            <Icon icon={faDatabase} />
+            &nbsp;&nbsp;{message('project.uploadBundle')}
+          </Button>
+        </Stack>
```
ansoncfit, line 60: Was it an intentional change to remove the conditional logic for displaying Create Project vs. Upload Bundle? https://github.com/conveyal/analysis-ui/blob/43afab8f57a403afed061c5bf93a88ad9b97637c/lib/components/select-project.js#L26

lib/components/variant-editor.tsx
```diff
@@ -0,0 +1,136 @@
...
+      <Stack isInline spacing={0}>
+        <IconButton
+          icon={faEye}
+          label={message('variant.showModifications')}
+          onClick={showVariant}
+        />
+        <IconButton icon={faCopy} label='Copy scenario' onClick={copyVariant} />
+        {index !== 0 && (
+          <IconButton
+            icon={faTrashAlt}
```
ansoncfit, line 127: Is there a reason scenarios use `faTrashAlt` and modifications use `faTrash`?

lib/components/variant-editor.tsx
```diff
@@ -0,0 +1,136 @@
...
+            label={message('variant.delete')}
+            onClick={onOpen}
+            variantColor='red'
```
ansoncfit, line 130: The red is eye-catching -- what would you think about making it lighter?

lib/components/variant-editor.tsx
```diff
@@ -0,0 +1,136 @@
...
+          onConfirm={deleteVariant}
+        />
+      )}
+
+      <Text mr={2}>{index + 1}.</Text>
+      <Box flex='1' fontWeight='bold'>
+        <Editable isValid={isValidName} onChange={onChangeName} value={name} />
+      </Box>
+      <Stack isInline spacing={0}>
+        <IconButton
```
ansoncfit, line 119: For consistency with the modifications list, I'd prefer the visibility icon for scenarios to be rightmost.

lib/hooks/use-route-to.ts
```diff
@@ -0,0 +1,25 @@
+import isEqual from 'lodash/isEqual'
+import Router from 'next/router'
+import {useCallback, useEffect, useState} from 'react'
+
+import {routeTo} from '../router'
```
ansoncfit, line 5: I could use a primer on how this router works. Ideally, after selecting a project on the single-point analysis page, navigating to the regional analysis page, and navigating back in the browser, the single-point analysis settings (including selected project) would be maintained. Currently, it looks like the projectId is registered as a query parameter, but it is not passed back to the single-point analysis panel.

lib/components/modification/editor.tsx
```diff
@@ -0,0 +1,351 @@
...
+        {...this.props}
+        debouncedSaveToServer={this.debouncedSaveToServer}
+        updateLocally={this.updateLocally}
+      />
+    )
+  }
+}
+
+/**
+ * Show this toast when a modification has been copied.
```
ansoncfit, line 100: 🍞

lib/components/modification/editor.tsx
```diff
@@ -0,0 +1,351 @@
...
+      cursor='pointer'
+      onClick={() => {
+        goToModificationEdit()
+        onClose()
+      }}
+    >
+      <AlertIcon />
+      <AlertDescription>
+        <PseudoBox _hover={{textDecoration: 'underline'}}>
+          Modification has been copied. Click to go to the newly created
```
ansoncfit, line 123: Maybe "Modification has been copied successfully. Click here to go to the new copy."

Nate-Wessel:
These changes look really nice! Big improvement. 

One thing I might suggest is that it would be nice in the scenario view to highlight the scenario, if any, that matches the map display. Clicking the eye icon gives it a little blue highlight as feedback, but that disappears after a couple seconds.

lib/components/select-project.tsx
trevorgerhardt: Ah, yes, I was experimenting with adding more region specific actions there (add an opportunity dataset) and then just went with what was already there.

lib/components/variant-editor.tsx
trevorgerhardt: It is possible, but it refers to a Chakra default so we would have to update the entire theme.

lib/components/variant-editor.tsx
trevorgerhardt: Nope, accidental change. Reverted.

trevorgerhardt:
> These changes look really nice! Big improvement.
> 
> One thing I might suggest is that it would be nice in the scenario view to highlight the scenario, if any, that matches the map display. Clicking the eye icon gives it a little blue highlight as feedback, but that disappears after a couple seconds.

I like that idea @Nate-Wessel but it would take a bit of logic to implement and I think would be good to do in a follow up PR. Can you create an Issue describing that?

abyrd:
I agree, looks very good. One thing I noticed: when you click the modification name to edit it, the font changes to something smaller. I guess that's to make space for the underline that indicate's it's being editable. But I always am a little bothered by editable text fields like browser URL displays that change location/content/size when you click them to edit.

lib/hooks/use-route-to.ts
ansoncfit: For a URL like `modifications/[modificationId]`, the nextjs router is responsible for looking up the page (modifications) and setting the state of the page (e.g. modificationId). Incremental rendering from nextjs makes the page transitions smooth.

Our own code (rather than the nextjs router) handles "after-the-question-mark" search parameters -- we'll need to investigate that to resolve #1147.

PR #1121, opened 2020-06-22
Building out cypress tests

Nate-Wessel:
Another iteration developing the Cypress tests, this one building out some tests related to running an actual analysis. 

Some of the big changes are:
* pending tests - I've outlined some future work with empty tests that display as "pending"
* taking a foray into visual snapshot testing - this has been inconsistent however between platforms, so at present the automatic test runner will not find any comparison snapshots and will try to replace them with new ones, allowing the tests to pass without actually doing the comparison. A few snapshots are stored for local tests while I try to figure this out. 
* I turned off the `cypress/no-unnecessary-waiting` rule in local/analysis.js. Many of the hard waits are related to elements that disappear from the DOM, indicating some event, but then return. They are no longer than necessary. 
* `fixtures/regions/scratch-results.json` contains numerical results from the analysis tests. 
* Many elements have been newly (re)labelled. Thanks @trevorgerhardt!


Nate-Wessel:
[Cypress-image-snapshot](https://github.com/palmerhq/cypress-image-snapshot) saves/searches for snapshots based on the context in which the snapshot function is run. This introduces problems when running just one spec file locally vs. the integrated `all.js`spec file. The snapshots would be saved in two different places. e.g.
`analysis-ui/cypress/snapshots/local/analysis.js/*.png`
vs
`analysis-ui/cypress/snapshots/local/all.js/*.png`

With `all.js` there is also greater potential for name collision as more snapshots are added to multiple tests and test files.

Nate-Wessel:
> [Cypress-image-snapshot](https://github.com/palmerhq/cypress-image-snapshot) saves/searches for snapshots based on the context in which the snapshot function is run. This introduces problems when running just one spec file locally vs. the integrated `all.js`spec file. The snapshots would be saved in two different places. e.g.
> `analysis-ui/cypress/snapshots/local/analysis.js/*.png`
> vs
> `analysis-ui/cypress/snapshots/local/all.js/*.png`
> 
> With `all.js` there is also greater potential for name collision as more snapshots are added to multiple tests and test files.

It looks like these image snapshots are not consistent between local and GitHub CI runs. When running locally they are consistent between headed and headless modes though, so that shouldn't be the issue. We still only have snapshots referring to one of the test modules (`analysis.js`), but not to `all.js`. Thus, the snapshot check will still not apply in the automatic CI tests, but may be used if running locally.

PR #1123, opened 2020-06-24
Add console error and exit if percentile curves do not exist

### LINKED ISSUE #1122: Boxplot without percentile curves is throwing an error
From: https://app.logrocket.com/7kj4kr/conveyal-analysis-production/s/4-ada77b2d-570e-4a66-8779-1807769dc3b1/0?stack_trace_id=707&stack_trace_tab=715bb3da-acd8-45fd-aaf9-60ae72e29d37&t=1592847471749.7432&undefined=WARN%252CERROR

It appears that the graph is being rendered when no data exist. Which should not be occuring.

trevorgerhardt:
Errors were occurring when the UI was attempting to render a Boxplot when data did not exist. I could not reproduce the problem locally, so this PR adds more protections for rendering a Boxplot and also logs an error in the case that it attempts to when it shouldn't.

Fixes #1122


trevorgerhardt:
@ansoncfit here is the commit I accidentally pushed to dev. Main thing done there was simplify/refactor the ErrorModal handling cleaning up it's inputs at a level higher. We now handle fetch errors specifically inside the `with-redux` handler and application errors (NPEs and such) in `pages/_app.tsx`. That's where they were coming from before, but we were formatting them inside the ErrorModal which made it complicated.

Also, it removes a `console.error` which I believe was creating a lot of LogRocket errors that are not 100% necessary -- for example when the Analysis request returns an error we show useful information to the user and it doesn't need to be logged as a LogRocket error.

https://github.com/conveyal/analysis-ui/commit/e27cdad9aa2b012e4974b41de85835a996ae923b

PR #1135, opened 2020-07-02
allow basemap style to be set with NEXT_PUBLIC_MAPBOX_STYLE

mattwigway:
## Description

Currently the frontend can't be run by anyone without a Conveyal-issued Mapbox key, because the default Mapbox style is private. This allows overriding the basemap style by setting the environment variable NEXT_PUBLIC_MAPBOX_STYLE to a fully-qualified mapbox style name (e.g. `mapbox/light-v10`).

I used the NEXT_PUBLIC_ prefix rather than edit `next.config.js` as this seems preferred for recent versions of next.

The environment variables don't seem to be documented anywhere, but happy to document this variable if some points me where to do it.

## How to test

1. Set `MAPBOX_ACCESS_TOKEN` to a token that isn't from Conveyal.
2. Load Analysis, note that it errors loading the basemap (basemap is just gray).
3. Set `NEXT_PUBLIC_MAPBOX_STYLE=mapbox/light-v10`, restart, and observe that the basemap now works.


mattwigway:
Okay, I rebased off `dev` (though I haven't tested since I don't have the backend running locally atm). I don't see a default `.env` file in the repo, where is that?

trevorgerhardt:
@mattwigway https://github.com/conveyal/analysis-ui/blob/dev/.env

mattwigway:
👍

lib/leaflet.js
```diff
@@ -1,14 +1,15 @@
...
 const MB_URL = 'https://api.mapbox.com/styles/v1/conveyal'
-const MB_STYLE = 'cjwu7oipd0bf41cqqv15huoim' // based off "Light"
+const MB_STYLE =
+  process.env.NEXT_PUBLIC_MAPBOX_STYLE || 'conveyal/cjwu7oipd0bf41cqqv15huoim' // based off "Light"
 
 /**
  * Allow passing a style to get the full Leaflet url.
  */
 export function getTileUrl(style = MB_STYLE) {
+  // TODO does not appear to be called anywhere (MWC 2020-07-02)
```
trevorgerhardt, line 12: If it's not called anywhere this can be removed.

PR #1143, opened 2020-07-09
Update User Manual

ansoncfit:
This PR adds "Learn More" sections for:
* Cycling LTS
* OSM license
* Park-and-ride access mode

It should not be merged to master until the recommended version of R5 is updated to a version that includes the park-and-ride tag changes.


docs/learn-more/lts.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Labeling Network Edges with LTS
+
+### Conveyal Surrogate LTS
+
+Conveyal Surrogate LTS is the default methodology in Conveyal Analysis for labeling network links based on commonly tagged attributes of OSM ways. The labeling logic in v5.9.0 of the routing engine can be viewed in its [source code](https://github.com/conveyal/analysis-backend/blob/v5.9.0/src/main/java/com/conveyal/r5/labeling/LevelOfTrafficStressLabeler.java) and is summarized below:
+* If way has tag highway:service, do not label edge
+* If way does not allow cars or has tags highway:residential or highway:living_street, label edge as LTS 1
+* If speed limit is less than 40.25 kph, and the number of lanes is less than 4 or unspecified, label edge as LTS 2
```
Nate-Wessel, line 19:
```suggestion
* If speed limit is less than 25mph (40.25 kph), and the number of lanes is less than 4 or unspecified, label edge as LTS 2
```
Assuming the audience is mostly Americans.

docs/learn-more/pnr.md
```diff
@@ -0,0 +1,5 @@
+# Park and Ride
+
+When transit modes are selected in analysis, park-and-ride can be selected as an access mode. Park-and-ride access uses locations identified based on certain OSM tags in the street layer of the baseline network, specifically areas and ways tagged with park_ride=* or parking=park_and_ride tags.
+
+This access mode mode requires the car to be parked and transit to be used before the journey is complete. Therefore the isochrones do not radiate out from the origin point; they radiate out from the destination transit stops that can be reached by parking the car, walking to a transit stop, riding a transit vehicle, then exiting transit.
```
Nate-Wessel, line 5:
```suggestion
This access mode requires the car to be parked and transit to be used before the journey is complete. Therefore the isochrones do not radiate out from the origin point; they radiate out from the destination transit stops that can be reached by parking the car, walking to a transit stop, riding a transit vehicle, then exiting transit.
```

docs/learn-more/lts.md
abyrd: It's likely at least half of readers will be outside the US, but many in former-British-empire places where they'll at least recognize miles or mph as units. In other places these units can be so unfamiliar as to be confusing. Generally I'd prefer putting the SI units first with imperial units in parentheses. Like 12 hour clocks and certain date formats imperial distance units can give the impression that we are parochially "nonstandard" and create vague concern that we don't understand how things work in most of the world. I'd rather err on the side of conventions of the places where public transit is heavily present, but provide imperial conversions for reader convenience.

However: in this case the kph value appears to be a fractional conversion of an integer number of mph. In that case there's a good reason to put the mph first.

docs/learn-more/pnr.md
```diff
@@ -0,0 +1,5 @@
+# Park and Ride
+
+When transit modes are selected in analysis, park-and-ride can be selected as an access mode. Park-and-ride access uses locations identified based on certain OSM tags in the street layer of the baseline network, specifically areas and ways tagged with park_ride=* or parking=park_and_ride tags.
```
abyrd, line 3: Actually by far the most common value for the tag park_ride is `no` (https://taginfo.openstreetmap.org/keys/park_ride#values). So we should probably say any park_ride value except `no`. I can make this change to the docs, then propose a PR to change the loading behavior in the software.

abyrd:
I have expanded and revised the park and ride information after an exchange with a user. Please review the new content before merging.

PR #1150, opened 2020-07-20
Adjust speed buttons (and other modification refactoring)

### LINKED ISSUE #1015: Selector tools' "remove from" overflows
The selector tools in the _remove stops_ and _adjust speed_ modifications both have an option to "Remove from" the current selection. In both cases it overflows it's div. 

Screenshot_2020-02-04_15-46-45

segment-selector

Perhaps these could be styled as flex-boxes?

trevorgerhardt:
General refactoring:
- Removes an additional "container" (we no longer use the Redux container/component differentiation)
- Converts more files to TypeScript
- Allow `confirm-button` handler to close (usually removes the element through navigation)
- Prefer `useState` + `useEffect` instead of `React.memo` for finer control, better TypeScript typing, and better display names
- Fixes scenario list scrolling and name overruns

Closes #1015, #1208, and #1234


lib/components/modification/adjust-speed.tsx
```diff
@@ -0,0 +1,141 @@
...
+      action: 'new'
+    })
+  }
+
+  function addToSelection() {
+    setMapState({
+      state: MAP_STATE_HOP_SELECTION,
+      action: 'add'
+    })
+  }
```
ansoncfit, line 58: The Add To mode only works if a segment has first been selected using the Select mode. Otherwise, an `Invalid attempt to spread non-iterable instance` error occurs. This behavior is present on dev, not just this branch.

trevorgerhardt: This has been fixed

trevorgerhardt:
This PR is code ready and ready for another review, but integration tests are failing and could use the Cypress changes introduced in #1199 to help pass.

lib/components/modifications-map/draw-polygon.js
```diff
@@ -33,7 +33,8 @@ export default function DrawPolygon(p) {
           polygon: true,
           rectangle: false,
           circle: false,
-          marker: false
+          marker: false,
+          circlemarker: false
```
ansoncfit, line 37: 👍

lib/components/modification/adjust-speed.tsx
```diff
@@ -0,0 +1,174 @@
...
+      {modification.hops && (
+        <Pane zIndex={230}>
+          <HopLayer
+            color={colors.MODIFIED}
+            feed={feed}
+            hopStops={hopStops}
+            modification={modification}
+          />
+        </Pane>
+      )}
```
ansoncfit, line 90: If `hops==null`, the speed factor applies to the entirety of the selected patterns, so the color should be `colors.MODIFIED`.

If `hops!=null` (i.e. only a subset of the patterns have speed adjusted), then there should be a NEUTRAL layer below the MODIFIED layer.

Does this code supersede `lib/components/modifications-map/adjust-speed-layer.js` LL 10-36?

lib/components/modification/adjust-speed.tsx
```diff
@@ -0,0 +1,174 @@
...
+              <IconButton
+                icon={faPlusSquare}
+                label={message('common.addTo')}
+                onClick={() => setAction('add')}
+                size='lg'
+              />
+              <IconButton
+                icon={faMinusSquare}
+                label={message('common.removeFrom')}
+                onClick={() => setAction('remove')}
```
ansoncfit, line 147: This action appears to have the opposite effect of what's intended. If a set of hops A-B-C is selected, drawing a polygon to remove subset A-B appears to remove subset B-C.

I'm not sure how commonly this remove-hops feature is used, so maybe we should just eliminate it.

lib/components/modification/adjust-speed.tsx
```diff
@@ -0,0 +1,174 @@
...
+      {get(modification, 'routes.length') === 1 && (
+        <Stack spacing={4}>
+          <SelectPatterns
+            onChange={(trips) => updateAndRetrieveFeedData({trips})}
+            trips={modification.trips}
+          />
+
+          <Alert status='info'>
+            <AlertIcon />
+            {message('report.adjustSpeed.selectInstructions')}
```
ansoncfit, line 124: This message should be updated, maybe with some conditional logic.

If `hops == null`, 

"Speed adjustment will apply to entire pattern(s). Use button below to limit the adjustment to certain segments."

else

"Speed adjustment will apply only to selected segments."

ansoncfit (changes requested):
Testing this out, the remove-stops updates look good. There's still a bit of confusing behavior/instructions with the adjust-speed modifications.

lib/components/modification/adjust-speed.tsx
trevorgerhardt: We had a distinction between the "editing" map components of a modification and non-editing map components. I moved the editing components directly into the individual modification components.

lib/components/modification/adjust-speed.tsx
trevorgerhardt: There was a note about the logic there. Corrected it. Works now! This problem exists on production too.

lib/components/modification/adjust-speed.tsx
ansoncfit: Just discussed: the code in `lib/components/modifications-map/adjust-speed-layer.js` is not duplicative -- it's separate logic for "view-only" display of the modification.

ansoncfit (approved):
Thanks for the fixes

PR #1155, opened 2020-07-26
Building out Cypress tests

Nate-Wessel:
This will expand coverage of the Cypress integration tests. Better description to come, when it's a bit further along. 

Added tests for:
* update region settings (full CRUD test)
* upload a broken GTFS feed and check for error messages
* upload multiple GTFS feeds

Fixed:
* problems with inconsistent result snapshots - stopped using today's date
* allowed snapshot testing of charts with 3% threshold, allowing tests to pass with ~1.5% cross-platform diffs

Etc.: 
* disabled video recording since they don't get saved anyway


trevorgerhardt:
@Nate-Wessel can we merge this? It contains the date change which fixes the analysis tests when run on weekends.

trevorgerhardt:
@Nate-Wessel if it's good to merge, just change it from a draft and it will merge.

PR #1156, opened 2020-07-27
Add the ability to select destination grids

trevorgerhardt:
Add a selector to allow selecting a destination point set in a regional analysis.
Additionally:
- Move the legend onto the map
- Refactor aggregation area upload to use Chakra components
- Allow downloading entire request JSON

Requires: https://github.com/conveyal/analysis-internal/pull/128


lib/components/analysis/profile-request-display.tsx
```diff
@@ -0,0 +1,234 @@
...
+const TDValue = ({children}) => (
+  <Box as='td' width='65%'>
+    {children}
+  </Box>
+)
+
+const PROJECT_CHANGE_NOTE =
+  'Notice: project may have changed since the analysis was run.'
+const SCENARIO_DOWNLOAD_NOTE =
+  'Download scenario info and modifications used to create this analysis.'
```
ansoncfit, line 66: `message` doesn't work for these?

lib/components/analysis/aggregate-accessibility.tsx
```diff
@@ -83,69 +84,74 @@ export default class AggregateAccessibilityComponent extends React.Component {
...
-            {' '}
-            {message('analysis.weightedAverageAccessibilityWarning')}
-          </span>
-        </div>
-      </>
-    )
-  }
+      <Alert status='warning'>
+        <AlertIcon />
+        {message('analysis.weightedAverageAccessibilityWarning')}
```
ansoncfit, line 151: This is detached from the weighted average statistic above, in a way that could be confusing.

Using a `Tooltip` (similar to the panel above for `PROJECT_CHANGE_NOTE`) might be preferable.

lib/components/analysis/aggregation-area.tsx
```diff
@@ -0,0 +1,189 @@
...
+      {!union && (
+        <Stack spacing={4}>
+          <Alert status='warning'>
+            A separate aggregation area will be created for each feature (up to
+            60) in the uploaded shapefile.
+          </Alert>
+          <FormControl isDisabled={uploading} isRequired>
+            <FormLabel htmlFor={attributeInput.id}>
+              Attribute name to lookup on the shapefile
+            </FormLabel>
```
ansoncfit, line 171: Prefer `message`?

lib/components/analysis/legend.tsx
```diff
@@ -0,0 +1,40 @@
+import {PseudoBox, Stack} from '@chakra-ui/core'
+import {format} from 'd3-format'
+import reverse from 'lodash/reverse'
+
+import {isLight} from 'lib/utils/rgb-color-contrast'
+
+const textFormat = format(',.0f')
+
+/** Display a map legend */
```
ansoncfit, line 9: Looking good.

lib/components/analysis/mode-summary.tsx
```diff
@@ -0,0 +1,47 @@
...
+        >
+          <Stack align='center' isInline spacing={1}>
+            <Box color={`${color}.500`} fontSize='xs'>
+              <Icon icon={faChevronRight} />
+            </Box>
+            {transitModesArray.slice(0, max).map((m) => (
+              <ModeIcon mode={m} key={m} />
+            ))}
+            {transitModesArray.length > max && <Box>...</Box>}
+            {egressModes !== 'WALK' && (
```
ansoncfit, line 34: Why hide the egress mode when it's walk?

trevorgerhardt: That's the common/assumed use case. We could do it for access mode too. But because we show walk as an access mode it's "known" that we are showing both access mode and transit modes inline.

PR #1166, opened 2020-08-06
Persist selected projectId across pages

trevorgerhardt:
- Persist the selected project across pages more consistently.
- Enable stricter integer parsing for Regional Analysis creation


lib/components/edit-bundle.tsx
```diff
@@ -164,7 +164,6 @@ export default function EditBundle(p) {
           ) : (
             <ConfirmButton
               action={message('bundle.delete')}
-              block
```
ansoncfit, line 167: What is the effect of moving this?

trevorgerhardt: It was unused, just noticed it when I was adding the id to the form input above.

PR #1167, opened 2020-08-06
Correct description of R5 bug

ansoncfit:


abyrd:
I just noticed we also have a section in the changelog saying "Allows transferring from one trip pattern to the same pattern, which may reduce travel times in systems with routes that loop or cross themselves." As discussed on our call, the bug we fixed was in the code that was intended to prevent reboarding. It was not succeeding at preventing boarding; it was instead preventing boarding an arbitrary route instead of the last route used. And in fact I think that arbitrary blocked route could not be the one that was just alighted from (unless walking from another stop on the route was faster than riding from that stop), so reboarding was probably always allowed and considered.

abyrd:
Please ignore my (now deleted) comment - I was looking at the text that was removed/corrected by this PR.

PR #1183, opened 2020-08-12
Don't show invalid travel time distributions

### LINKED ISSUE #1086: Strange travel times reported outside of region
Analysis reports travel time ranges outside the defined region and well outside the the transit/street network. It's not clear what the times are based on and sometimes they are rendered strangely or are surprisingly low. 

Screenshot_2020-05-06_10-03-40

Screenshot_2020-05-06_10-13-08

I would expect something like this where travel times are seemingly off the chart:
Screenshot_2020-05-06_10-07-41

Just a small distance away though, there are travel times again. 
Screenshot_2020-05-06_10-09-17

This is using just the baseline scenario and is present in the production version.

trevorgerhardt:
Invalid/garbage values were being shown for travel time distributions outside of the returned area. This fixes the issue by returning 0 if the values requested are out of bounds.

Also 
- Moves the graph from a popup to a static map overlay (similar to the regional analysis legend). 
- Displays the minutes corresponding to the selected percentile next to plot
- Shows the coordinates of the position
- Allows pinning the position by clicking on the map, and unpinning by clicking on the marker.

closes #1086 

## How to test

1. Check for a destination travel time distribution outside of the analysis bounds.


trevorgerhardt:
I think we can! I will try that

trevorgerhardt:
@ansoncfit works, and that looks pretty cool. But is there a need to keep the destination in one place?

lib/components/map/destination-travel-time-distribution.tsx
```diff
@@ -1,80 +1,231 @@
...
+              {comparisonDistribution &&
+                isValidTime(comparisonDistribution[percentileIndex]) && (
+                  <Text
+                    color='red.500'
+                    fontSize={FONT_SIZE}
+                    fontWeight='bolder'
+                    position='absolute'
+                    top={HEIGHT + 'px'}
+                  >
+                    {distribution[percentileIndex]}
```
ansoncfit, line 182: `comparisonDistrubition[percentileIndex]` ?

lib/components/map/destination-travel-time-distribution.tsx
```diff
@@ -1,80 +1,231 @@
...
+                        width={HEIGHT}
+                      />
+                    </g>
+                  )}
+                </svg>
+              </figure>
+            </Box>
+          ) : (
+            <Alert status='info'>
+              <AlertIcon /> Mouse over the isochrone to show travel times.
```
ansoncfit, line 224: Should be hidden if no isochrone is displayed (or switched to instructions for fetching results)

lib/components/analysis/boxplot.tsx
```diff
@@ -14,11 +15,11 @@ export default memo<BoxplotProps>(function Boxplot({
...
   scale,
+  strokeWidth = 0.5,
   width
 }) {
   const [low, iqrLow, med, iqrHigh, high] = positions
   const stroke = color
-  const strokeWidth = 0.5
 
   const center = 0.5 * width
   const whiskerLeft = 0.2 * width
```
ansoncfit, line 25: Feature request (could be ignored in this iteration): add a 2 * strokeWidth line at the selected percentile. This would make it much easier to understand what the displayed numbers to the left of the boxplot represent.

lib/components/map/destination-travel-time-distribution.tsx
```diff
@@ -1,80 +1,231 @@
...
-                <Boxplot
-                  color={colors.COMPARISON_PERCENTILE_COLOR}
-                  positions={comparisonDistribution}
-                  scale={SCALE}
-                  width={BOXPLOT_HEIGHT}
-                />
-              </g>
+    <>
+      {latlng && (
+        <CircleMarker
```
ansoncfit, line 143: Is an explicit layer order needed here to ensure the destination marker is clickable over isochrones and modification geometries?

trevorgerhardt: Are there instances where it's not clickable for you? Ordering is difficult....

lib/components/analysis/boxplot.tsx
trevorgerhardt: Add this note to the Issue for highlighting the cumulative accessibility plot?

lib/components/map/destination-travel-time-distribution.tsx
trevorgerhardt: Whoops

trevorgerhardt: Fixed

lib/components/map/destination-travel-time-distribution.tsx
trevorgerhardt: Updated to hide this until an isochrone is available.

lib/components/map/destination-travel-time-distribution.tsx
ansoncfit: I thought I noticed intermittent cases where it was not clickable, but I can't reproduce them now.

PR #1210, opened 2020-08-28
Enable worker-side accessibility calculation by specifying opportunities

abyrd:
This is based on the changes we worked through together in today's work session.
When an opportunity dataset is selected in the UI send its ID (in an array).
I factored out an ID selector from the activeOpportunityDataset selector.


trevorgerhardt:
@abyrd I pushed changes enabling using the results sent from the server. As you can see, this change set stayed quite small to enable this new behavior. Since we already have a slate of changes ready I think we should push off moving the Opportunities selector into the individual primary and comparison request forms until later since that may take some design iterations. We can do that along with the discussed custom JSON editor in a subsequent pull request.

abyrd:
Thanks for the additional code to handle the accessibility values received by the UI. It's great to see this can be done simply - that's what I was hoping for. I agree we should get this functionality into the UI in its simple form, and make those other changes (moving opportunities selector into request settings, custom JSON editor) later. I will remove the "draft" status from this, feel free to merge if it's working in testing (I will test it out tomorrow).

trevorgerhardt (approved):

PR #1218, opened 2020-09-06
Convert new APIs to handle all collection types

trevorgerhardt:
With the "catch all" route feature provided by Next.js we do not need duplicate code for each collection that falls under our standard "accessGroup" strategy. This PR converts the endpoints created for regions into generic `[collectionName]` endpoints allowing us to query `/api/db/regions` like before and also `/api/db/projects` or any other collection without adding new endpoints or models.

We can be certain that this is safe, because all model's/collections use the same permissions checking.


ansoncfit (approved):
Approving for merge into the `gql` branch

PR #1227, opened 2020-09-08
Presets (previously known as Bookmarks)

### LINKED ISSUE #638: Allow deleting bookmarks
I accidentally created a bookmark with the origin at a bar, and access mode as car, and would like to delete this bookmark (🍷  🚫   🚗) 

But in seriousness, a client has requested the ability to delete bookmarks.

trevorgerhardt:
This PR enables preset management. Deleting, editing names, filtering, and viewing the properties of the presets.

It should work on staging for new presets, but will not show old bookmarks until we do the migration.

closes #638


abyrd:
I'm reviewing this but having some trouble interpreting what I'm seeing. Several files are added or changed in `lib/components/modification/`, can you explain briefly how this is connected to saving presets? Also, am I correct in understanding that preset storage is using the new catch-all `/api/db/x` endpoints, and if so can you point out where (in what file) that API call would be made?

lib/hooks/use-model.ts
```diff
@@ -8,34 +8,43 @@ import {UserContext} from '../user'
...
       ...results,
       update,
-      remove: () => safeDelete(`/api/db/${collectionName}/${_id}`)
+      remove
     }
   }
 }
 
 export const useRegion = createUseModel('regions')
+export const useAnalysisPreset = createUseModel('analysisPreset')
```
trevorgerhardt, line 50: This needs to be corrected to `usePreset = createUseModel('presets')`

trevorgerhardt:
> I'm reviewing this but having some trouble interpreting what I'm seeing. Several files are added or changed in `lib/components/modification/`, can you explain briefly how this is connected to saving presets? Also, am I correct in understanding that preset storage is using the new catch-all `/api/db/x` endpoints, and if so can you point out where (in what file) that API call would be made?

Are you accidentally looking at a merge commit? There aren't any files changed under `/lib/components/modification` in this PR specifically.

The preset storage is using the new catch-all api routes. In order to use them, I created helper "hooks" for CRUD actions. See `lib/hooks/use-model` and `lib/hooks/use-collection`. This both have simple factory functions that encapsulate the collection name, so we have a `usePreset(id)` hook created in `use-model` and a `usePresets()` hook created in `use-collection`.

Each additional collection we can just add a line to the `use-model` file and `use-collection` file like:
```tsx
// use-model
export const useProject = createUseModel('projects')

// use-collection
export const useProjects = createUseCollection('projects')
```

migrations/1599703681398-bookmarks-to-presets.js
```diff
@@ -0,0 +1,12 @@
...
+module.exports.up = handler(async function (db) {
+  await db.renameCollection('bookmarks', 'presets')
+})
+
+module.exports.down = handler(async function (db) {
+  const collections = await db.collections()
+  if (collections.find((c) => c.collectionName === 'presets') != null) {
+    await db.renameCollection('presets', 'bookmarks')
+  }
+})
```
ansoncfit, line 12: Looks like a pretty straightforward migration.

lib/components/analysis/egress-mode-selector.js
```diff
@@ -1,42 +0,0 @@
-import {faBicycle, faWalking} from '@fortawesome/free-solid-svg-icons'
```
ansoncfit, line 1: With egress mode now handled in the `ModeSelector` component, this is unused 👍

ansoncfit (changes requested):
Great to see these updates and the ability to rename/delete presets. A few cleanup items:

- It appears that creating a preset before the origin marker has been moved (i.e. when it defaults to the center of the region) saves the preset without `fromLat` and `fromLon` values. This means once the marker is moved, loading (and by extension, editing or deleting) the preset is not possible unless the user manually moves the origin back to the initial location exactly.

- Working through this behavior, it occurred to me that we could make the relocation of the origin marker to a preset's saved `fromLat`/`fromLon` optional, either by giving the option to include origin coordinates when saving a preset (and not resetting the origin location for presets saved without origin coordinates), or by adding a move origin to preset location button with the preset options. That's additional functionality, so we could table it.

- Logic for the "Identical request settings" switch in the comparison request should be checked. With the current implementation, if a preset is loaded for the comparison request, it can have different settings than the main analysis but still show up with "identical request settings" true.

lib/components/presets.tsx
```diff
@@ -0,0 +1,350 @@
...
+        position: 'top',
+        status: 'error',
+        duration: null,
+        isClosable: true
+      })
+    }
+  }
+
+  return (
+    <Modal isOpen={true} onClose={onClose} initialFocusRef={nameInput.ref}>
```
ansoncfit, line 259: `closeOnOverlayClick={false}` here and in the other added Modals?

lib/auth0.ts
```diff
@@ -8,9 +8,11 @@ import {IUser} from './user'
 const cookieLifetime = ms('30 days') / 1000
 const httpTimeout = ms('10s')
 const scope = 'openid profile id_token'
-const redirectUri = `https://${process.env.VERCEL_URL}`
 
-const auth0 = (function createAuth0() {
+// Initialzed auth0s per origin served from this lambda
```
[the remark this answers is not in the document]
trevorgerhardt: Updated

lib/components/presets.tsx
```diff
@@ -0,0 +1,350 @@
...
+  // Remove a preset and show a toast on success
+  const _removePreset = useCallback(
+    async (_id) => {
+      const res = await presetsCollection.remove(_id)
+      if (res.ok) {
+        toast({
+          title: 'Deleted selected preset',
+          position: 'top',
+          status: 'success'
+        })
```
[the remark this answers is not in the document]
trevorgerhardt: Yea, another pull request addressed an issue with the confirm dialog that changed this functionality.

lib/components/presets.tsx
trevorgerhardt: Added 👍

trevorgerhardt:
> * It appears that creating a preset before the origin marker has been moved (i.e. when it defaults to the center of the region) saves the preset without `fromLat` and `fromLon` values. This means once the marker is moved, loading (and by extension, editing or deleting) the preset is not possible unless the user manually moves the origin back to the initial location exactly.

> * Working through this behavior, it occurred to me that we could make the relocation of the origin marker to a preset's saved `fromLat`/`fromLon` optional, either by giving the option to include origin coordinates when saving a preset (and not resetting the origin location for presets saved without origin coordinates), or by adding a move origin to preset location button with the preset options. That's additional functionality, so we could table it.

Hmm...ok, this is definitely a problem. I am more inclined to go with one default than add additional options. I think it would be clearer to always save the origin and always load the preset to the origin. If a user wants to delete/edit a preset they can always reselect it which would move it back to the correct origin.

> * Logic for the "Identical request settings" switch in the comparison request should be checked. With the current implementation, if a preset is loaded for the comparison request, it can have different settings than the main analysis but still show up with "identical request settings" true.

Fixed 👍

trevorgerhardt:
With the full separation we could even add an additional origin marker for the comparison.

ansoncfit (approved):
Thanks for the additional changes.

PR #1235, opened 2020-09-15
Disable recording to speed up cypress tests

trevorgerhardt:
Turns out we overused the free tier for Cypress. This PR updates Cypress to 5.2 and disables recording (which speeds up the tests in the Github Actions themselves).

### DIFF
--- a/.github/workflows/lint-jest-build.yml
+++ b/.github/workflows/lint-jest-build.yml
@@ -120,11 +120,11 @@ jobs:
         run: curl https://r5-builds.s3-eu-west-1.amazonaws.com/${{ matrix.backend }}.jar --output ${{ github.workspace }}/${{ matrix.backend }}.jar
 
       - uses: cypress-io/github-action@v2
-        env:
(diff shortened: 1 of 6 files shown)


.github/workflows/lint-jest-build.yml
```diff
@@ -120,14 +120,14 @@ jobs:
...
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # env:
+        # pass the Dashboard record key as an environment variable
+        # CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+        # pass GitHub token to allow accurately detecting a build vs a re-run build
+        # GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           install: false
-          record: true
+          # record: true
```
abyrd, line 130: Although we can guess from context that the default is probably false, might it be better to explicitly set `record: false` instead of commenting out `record: true` and relying on the default?

abyrd (approved):

PR #1243, opened 2020-09-21
Fix bug causing new modifications to not show on map

### LINKED ISSUE #1217: Modification visibility on map cannot be toggled
**Describe the bug**
After a modification is drawn or edited, its visibility icon in the modification list only works if the modification was exited by clicking on the back arrow at the top of the panel. If the user navigates away from the modification another way (e.g. clicking on the analysis tab), the modification won't be shown on the map even if the visibility icon for it or its scenario is clicked.

**To Reproduce**

1. Draw an add trip modification
2. Click on the analyze tab
3. Click on the modifications tab
4. Click the visibility icon to attempt to show the added route

trevorgerhardt:
closes #1217


ansoncfit (approved):
To recap our discussion:
There are a few simplifying changes here (e.g. not using a Set).

The core problem was related to `DisplayAll` component being used in two ways -- to display other modifications in background when editing one modification, and to display all modifications when viewing the modifications list. It exposes two different versions of itself: one that does not take props ("connected component," connected directly to redux store) and one that does.

modificationsOnMap is stored in local storage, per project. By default, updates to this map will not update components.

PR #1246, opened 2020-09-22
Try CodeQL

trevorgerhardt:
This add's a Github Action that run's Github's CodeQL code scanning on the JavaScript code in this repository. It's free for public repositories and I'm curious what it may uncover. This was an autogenerated file that I only changed to dry run for this PR. It should not have any other side effects as it's just an additional Github workflow file.

Learn more here: https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/managing-alerts-from-code-scanning


ansoncfit (approved):
I'll look forward to trying out the reports on merges and at 9 on Tuesdays.

Additional change to handle display of regional results when project/bundle have been deleted looks good.

Issue #628, opened 2018-01-10
Cannot render accessibility indicators with very few distinct values


abyrd:
This occurs especially when comparing two sets of results with almost no differences. If we ask for 5 colors and there are less than 5 distinct accessibility (accessibility difference) values it fails:
```
Error: cannot generate more classes than there are data values
    at ckmeans (ckmeans.js:220)
    at ckmeans.js:39
    at next (regional.js:95)
    at fetch.js:285
    at fetch.js:205
    at <anonymous>
```

abyrd:
Some insights: 

The ckmeans clustering is performed by the SimpleStatistics library. It would be nice if it would just produce less categories when it didn't have enough distinct values to cluster, be we can't directly change its behavior. 

According to https://simplestatistics.org/docs/#ckmeans it will throw an error in this situation. We need to catch the error and provide alternate behavior.

```
  try {
      const clusters = ssCkmeans(filtered, nBreaks)      
  } catch (error) {
      // Fallback: return one break per value
  }
```

We need to make sure this is the only situation in which an error is thrown, or somehow check the error type. We then need to find all the distinct values in the original or filtered data and make one category per value, or one category for each of the first N values and one catch-all for all the rest of the values.

We can see where these breaks are used in ChoroplethColorizer, it should accept variable length breaks but expects the array of colors to be the same length.

ansoncfit:
Closed by https://github.com/conveyal/analysis-ui/pull/1247

PR #1247, opened 2020-09-22
Prevent CKMeans throwing an error from lack of data

### LINKED ISSUE #1153: CKMeans "cannot generate more classes than there are data values"
When comparing analyses with identical results, sometimes there are no differences. Our classifier / color range generator cannot deal with this and we get an error. For a while it seemed like this problem was caught or avoided, but I have seen it again recently. It only shows up on certain comparisons - sometimes a comparison has no differences and it just doesn't display any legend or colors. Other times we get an error popup saying there are not enough values to classify.

Maybe there's a special case for "no difference at all" (only one value, zero) but no handling of cases with 3-4 values?

When there are no differences at all to display, it would be helpful to show this information (or somehow continue to display a legend with only zero and a solid color).

trevorgerhardt:
If an opportunity dataset or regional grid contains on data an error would be thrown when generating the map view. This PR adds checks that should prevent those "no data" errors.

closes #1153


ansoncfit (approved):
Should also close https://github.com/conveyal/analysis-ui/issues/628

PR #1251, opened 2020-09-24
Decay function selector & custom json editor

### LINKED ISSUE #930: New fields not updated in custom profile request JSON
Using branches with our freeform pointset changes, I appended this to the custom JSON and started a regional analysis:
```
"originPointSetId":"[uuid]",
"destinationPointSetId":"[uuid]",
"recordTimes": true
```

This triggered an error on the back-end because methods to record accessibility counts are not yet implemented.  I then edited the custom JSON to 

```
"originPointSetId":"[uuid]",
"destinationPointSetId":"[uuid]",
"recordTimes": true,
"recordAccessibility": false
```

After clicking "update profile request" and starting an analysis, the back-end showed the same error. Intercepting the request that the front-end sent, I noticed the "recordAccessibility" property was not included.

To confirm, I then used browser dev tools to clear the application's saved profile request and re-appended the custom JSON above, and the regional analysis completed successfully.

### LINKED ISSUE #1124: Add tooltip/explanation when multi-point button is disabled
The regional/"multi-point" button is disabled unless an isochrone is displayed; checking single-origin results is a verification step that helps avoid heavy computation for analyses with invalid settings.

I think we used to have a clearer explanation of this in the UI -- I handled two support requests about it today. We could hide the multi-point button, or add an explanation like "Fetch results for single point first. Multi-point analysis is disabled unless an isochrone is displayed."

### LINKED ISSUE #1130: Custom profile request JSON retains deleted fields
If a property is deleted from the custom profile request JSON in the UI, it is apparently not deleted from the stored version. The deleted property reappears in the JSON text box. This can get messy when editing the name of a property, for example after making a typo. The property may be saved in several intermediate states as you edit its name, creating n copies of the property with slightly different names.

### LINKED ISSUE #1249: Accessibility plot not invalidated when opportunities change or comparison project removed
In the single point analysis page, when using worker-side accessibility computations, if you change the destination set the cumulative accessibility plot is no longer reflecting parameters. Following existing convention it should be grayed out (along with some other controls) until the analysis is re-run.

Something similar is true for comparing two requests: when you remove the comparison project (by clicking the X) the plot is not grayed out. If you change the project or scenario rather than clear them, the plot is grayed out as expected.

### LINKED ISSUE #1253: Missing map layers
Opportunity dataset dot-density layers, and regional analysis accessibility layers, aren't showing up where expected.

trevorgerhardt:
### Changes
- Move Analysis custom JSON editor into a tab similar to modifications
- Add decay function selector and decay parameter editor
- Compact Analysis form
- Replace settings instead of merge settings when editing the JSON directly (closes #1130 and closes #930)
- Fix bug causing Gridualizer layers not to render (closes #1253)
- Misc. TypeScript migrations
- Add title to disabled regional analysis creation button (closes #1124)
- Show "analysis out of date" when accessibility has been calculated server side and then opportunities have changed (closes #1249)


abyrd:
That's a nice batch of changes. First impressions from testing this out a bit: the full sized JSON editor is great! And it's very nice to be able to select the decay function from the UI. When switching to the JSON it looks like the decayFunction property is accumulating config values for all the different function types, not just the currently selected one:
```
"decayFunction": {
    "type": "logistic",
    "standardDeviationMinutes": 10,
    "decayConstant": -0.01,
    "decayWidth": 10
  },
```

With the latest commit it does look like #1249 is solved for worker-side accessibility calculation (curves grayed out until the request is re-issued manually). I will continue testing and complete the review soon.

ansoncfit:
I agree the new JSON editor is a great change.

One issue - moving the time cutoff up to 120 minutes yields NaN in the access progress bars.

trevorgerhardt:
> That's a nice batch of changes. First impressions from testing this out a bit: the full sized JSON editor is great! And it's very nice to be able to select the decay function from the UI. When switching to the JSON it looks like the decayFunction property is accumulating config values for all the different function types, not just the currently selected one:
> 
> ```
> 
> "decayFunction": {
> 
>     "type": "logistic",
> 
>     "standardDeviationMinutes": 10,
> 
>     "decayConstant": -0.01,
> 
>     "decayWidth": 10
> 
>   },
> 
> ```
> 
> 
> 
> With the latest commit it does look like #1249 is solved for worker-side accessibility calculation (curves grayed out until the request is re-issued manually). I will continue testing and complete the review soon.
> 
> 
> 
> 

@abyrd is it ok to have all of the decay values set and just have the type change? When I was experimenting with them, changing them back and forth it would reset the values if I didn't just save them. Figured it was harmless to send unused parameters.

trevorgerhardt:
> I agree the new JSON editor is a great change.
> 
> One issue - moving the time cutoff up to 120 minutes yields NaN in the access progress bars.

Off by 1 error. I'm curious of the results now between access calculated on the server vs in the browser. Fix is here:

https://github.com/conveyal/analysis-ui/pull/1251/commits/f10de64cc7d9cd139fc83d315cc7b0c702fb9628#diff-04292ca1710fb9a3ee4fa402a599f88aL28-L33

ansoncfit (commented):
A few additional notes/suggestions:
- Resolve this error, which occurs when after fetching results with a comparison project set, clearing the comparison project, then re-fetching results:
> ```
> TypeError: Cannot read property 'projectId' of undefined
> The error may be correlated with this previous error:
> TypeError: Cannot read property 'projectId' of undefined
> ```
- Indicate stale results more consistently: when the opportunity dataset is changed, the cumulative plot is grayed out to indicate the results are stale. The stats above the plot should similarly indicate the result is stale, with a grayed out progress bar (and possibly a strikeout of the accompanying number)
- Consider replacing "Time cutoff" with "Travel time" (pending discussion)
- Ensure consistency between the selected decay function and the results shown in the plot. There are a couple cases when the plot will show non-stale, step-function results even when the selector is set to another decay function:
  - When R5 < v6 is used. To avoid this, we could locking the decay function selector to "step" for old versions. That feels a bit messy, but there's already precedent in using the R5 version number to enable different options in the regional analysis creation options.
  - When no opportunity dataset is selected, results are fetched, then an opportunity dataset is selected. To avoid this (and the case above), the decay selector could be reset to "step" anytime a single-point result with a null accessibility surface is returned. While this approach would avoid inconsistency, it has the downside of potentially modifying an input field, which could be hidden if the analysis settings are collapsed.

abyrd:
> @abyrd is it ok to have all of the decay values set and just have the type change? When I was experimenting with them, changing them back and forth it would reset the values if I didn't just save them. Figured it was harmless to send unused parameters.

It does work - the JSON deserialization is set to ignore values it doesn't recognize. I can see it being potentially confusing when debugging or examining the system, and it seems unconventional, or like "bad form". But if it is much simpler to implement by just sending all the parameters, pragmatically speaking it should work.

abyrd:
Using this latest commit 9561dd1, I see the following behavior: if I do a single point analysis with no destination pointset selected, using a non-step decay function, then after that result is displayed I select a destination pointset, a cumulative opportunities plot is shown without resending a single point request. Presumably this is done by combining the travel times client-side, which does not apply the decay function.

abyrd:
Testing this now with backend version `d44feed`, worker version `v5.10.0-28-g715ddc5` (which expects positive decay constants for fixed-exponential decay). Of course in its current form the UI can't send positive decay constants. It validates them and shows the red box indicating they are bad, as expected. But I notice it does let you trigger a request anyway despite the presence of inputs that failed validation. This sends a previous version of the request. Should the "fetch results" button be unavailable when inputs are failing validation?

abyrd:
Something still seems to be amiss with the interaction between decay functions and stale results. If I do a search with a step decay function, it displays results. Then I switch to a linear decay function with 10 minute width. Displayed results are invalidated (as expected). When I click "fetch results" the smoother results from the linear decay are displayed, but they are still grayed out with the text "results are out of sync with settings". Re-fetching multiple times continues to give the same results.

trevorgerhardt:
> Testing this now with backend version `d44feed`, worker version `v5.10.0-28-g715ddc5` (which expects positive decay constants for fixed-exponential decay). Of course in its current form the UI can't send positive decay constants. It validates them and shows the red box indicating they are bad, as expected. But I notice it does let you trigger a request anyway despite the presence of inputs that failed validation. This sends a previous version of the request. Should the "fetch results" button be unavailable when inputs are failing validation?

Overall, I think that is a good idea worth exploring. But it's existing behavior (with other input elements) and would be a big enough change that we should push off for now.

trevorgerhardt:
> Something still seems to be amiss with the interaction between decay functions and stale results. If I do a search with a step decay function, it displays results. Then I switch to a linear decay function with 10 minute width. Displayed results are invalidated (as expected). When I click "fetch results" the smoother results from the linear decay are displayed, but they are still grayed out with the text "results are out of sync with settings". Re-fetching multiple times continues to give the same grayed out results and message.

Ah yea, my first pass had the correct logic, but then subsequent clean up screwed it up. Fixed.

abyrd (approved):
All issues discussed have been addressed in new commits. We have decided to merge optimistically and will resolve any problems we might find via testing in future PRs.

Issue #31, opened 2016-03-28
add stops modification


mattwigway:
We want to create an add-stops modification, which will basically be a slice operator. It will have start and end stop IDs and will replace anything between them with a new routing.

mattwigway:
In the export the type is 'add-stop' but should be 'add-stops'

PR #1255, opened 2020-09-27
Improve handling of failed fetches

trevorgerhardt:
Failed fetches when the server is down or a connection is bad currently do not give very clear errors. In seeking to show more clarity there I discovered a mismatch in the logic of how we determine when fetches should throw on failure which would cause all failed fetches to just return void responses leading to errors further down, usually while rendering.

This fixes that error, and shows a new message specific for fetch failures.

Most of the files changed here are merely pushing off "destructuring" the returned response from a Promise.all, which if it contains an error, would fail to destructure and show that JavaScript error instead.

Also converted many of the files changed to TypeScript.


abyrd:
Testing the Vercel deployment of this, against the currently running staging server (version d44feed), when switching to the "resources" tab (currently a hidden feature) the fetch seems to fail with a 404 giving the following result:

I first get this popup:

```
Application error
Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=SyntaxError%3A%20Unexpected%20token%20%3C%20in%20JSON%20at%20position%200&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
```

When I close that one, I get another popup:

```
Error while communicating with the server
404 Not Found Unexpected token < in JSON at position 0
https://staging-api.conveyal.com/api/resources?projectId=5f2dfadf0ab36a325a55002c&regionId=5f28a6f8aef65d3daad0d2c3
```

This is due to receiving `<html><body><h2>404 Not found</h2></body></html>`. Questions are:

1. Is this failure expected when switching to the resources tab?
2. When such a failure happens is it expected that we get two popups in a row, one containing the decoded error, after the one containing the encoded error?
3. Should we be reacting to the 404 code to show an appropriate message instead of a general purpose message triggered by the body containing HTML instead of JSON?

I also notice that neither closing the error popups, nor using the "go back" button has the least-surprise effect of clearing away the errors. It's hard to find a sequence of actions to clear away the errors that feels stable and reliable.

abyrd:
Continuing testing as described above, and probably more in line with the intended scope of the PR: If I go to edit modifications, then block my internet connection before selecting a modification, I get an error displayed in two ways simultaneously. Once in the modal window ("Error while communicating with the server. Refreshing your browser may resolve the issue.") and once embedded in the underlying page (red box with "Error loading data. Failed to contact server. Please make sure you have an active internet connection.") 

Is this the intended result? In this case it seems like the red box would fully communicate the message. Are there other cases where a fetch would fail, but only the popup would be displayed and there's no red box?

trevorgerhardt:
> Continuing testing as described above, and probably more in line with the intended scope of the PR: If I go to edit modifications, then block my internet connection before selecting a modification, I get an error displayed in two ways simultaneously. Once in the modal window ("Error while communicating with the server. Refreshing your browser may resolve the issue.") and once embedded in the underlying page (red box with "Error loading data. Failed to contact server. Please make sure you have an active internet connection.")
> 
> Is this the intended result? In this case it seems like the red box would fully communicate the message. Are there other cases where a fetch would fail, but only the popup would be displayed and there's no red box?

I've updated this flow. 
- The red box in the side bar will appear if there is an issue on initial page load. 
- New "toasts" will show up if there is an uncaught issue in a subsequent server interaction.  
- The same modal with explanation will appear for Single Point Analysis errors.
- Each "error" is saved to LogRocket.

trevorgerhardt:
> Testing the Vercel deployment of this, against the currently running staging server (version d44feed), when switching to the "resources" tab (currently a hidden feature) the fetch seems to fail with a 404 giving the following result:
> 
> I first get this popup:
> 
> ```
> Application error
> Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=SyntaxError%3A%20Unexpected%20token%20%3C%20in%20JSON%20at%20position%200&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
> ```
> 
> When I close that one, I get another popup:
> 
> ```
> Error while communicating with the server
> 404 Not Found Unexpected token < in JSON at position 0
> https://staging-api.conveyal.com/api/resources?projectId=5f2dfadf0ab36a325a55002c&regionId=5f28a6f8aef65d3daad0d2c3
> ```
> 
> This is due to receiving `<html><body><h2>404 Not found</h2></body></html>`. Questions are:
> 
> 1. Is this failure expected when switching to the resources tab?
> 2. When such a failure happens is it expected that we get two popups in a row, one containing the decoded error, after the one containing the encoded error?
> 3. Should we be reacting to the 404 code to show an appropriate message instead of a general purpose message triggered by the body containing HTML instead of JSON?
> 
> I also notice that neither closing the error popups, nor using the "go back" button has the least-surprise effect of clearing away the errors. It's hard to find a sequence of actions to clear away the errors that feels stable and reliable.

The issues brought up here should be resolved with my latest changes. While the resources end point still does not work, these errors are all handled better now.

abyrd (approved):
Issues have been addressed in new commits, we have decided to merge optimistically and will revisit if we encounter any problems.

PR #1261, opened 2020-09-28
Remove old inputs and reduce CSS 

trevorgerhardt:
- Convert old inputs to Chakra inputs
- Convert old buttons to Chakra buttons
- Remove Bootstrap
- Clean up styles that used Bootstrap classes
- Remove unused CSS
- Migrate changed files to TypeScript


abyrd:
We have a chain of merges here from this #1261 into branch `failed-fetch` #1255, and from there into the decay function PR #1251. It's helpful to have these changes broken out for review in separate PRs, and I have done a quick once-over looking at the changeset but would mainly want to test them in use on staging. I'd prefer to optimistically merge these and then test everything together on staging as part of the decay function PR #1251. This should catch any problems due to possible interactions between changes.

@trevorgerhardt let me know if this approach works for you, and I'll review and approve with that path in mind.

trevorgerhardt:
> We have a chain of merges here from this #1261 into branch `failed-fetch` #1255, and from there into the decay function PR #1251. It's helpful to have these changes broken out for review in separate PRs, and I have done a quick once-over looking at the changeset but would mainly want to test them in use on staging. I'd prefer to optimistically merge these and then test everything together on staging as part of the decay function PR #1251. This should catch any problems due to possible interactions between changes.

That works. I wouldn't want to go through full testing before getting these last two PRs in, since the surface area is high, but the feature changes are low.

abyrd (approved):
It's hard to review the js -> tsx conversions since Github shows them as full deletions + new files, rather than rename + edit. It seems to me that command line and local tools do a better job of detecting/displaying renames. Anyway, overall looks good and we will do full testing on this once merged into the other PRs.

PR #1272, opened 2020-10-05
Only display unique error toasts

### LINKED ISSUE #1266: Toasts accumulate when UI cannot contact backend
This happens when staging shuts down; not sure if it would happen under any normal circumstances for end-users:

image

trevorgerhardt:
This change tracks if an error toast with the same title/description is already open and if it is, it will not show it again. 

closes #1266


ansoncfit (approved):
👍

PR #1274, opened 2020-10-05
Make modes selected more obvious

trevorgerhardt:
Using distinct styles instead of the button's "active" mode.

Screen Shot 2020-10-06 at 07 52 53


ansoncfit (approved):
👻 mode -- first step toward Conveyal's October 2020 theme.

PR #1275, opened 2020-10-06
Clarify create bundle

trevorgerhardt:
- Move info alert to the top and clarify that it's for after you click "Create"
- Add `isRequired` to File inputs
- Start on upload tabs, and disable reuse tabs, if no bundles exist
- Expand descriptions
- Use `DocsLink` for link to docs

Screen Shot 2020-10-06 at 08 36 50


i18n/en.json
```diff
@@ -53,8 +53,8 @@
     "processing": "Wait until processing is complete before leaving this page. This may take a few minutes.",
     "save": "Save renamed bundle/feed names",
     "select": "or select an existing one",
-    "notice": "You will need to wait a few minutes for new OSM and GTFS feeds to be processed.",
-    "createDescription": "When creating a new Network Bundle you have the option of reusing data from previously created Network Bundles. Note: Network Bundles created before April 15th, 2020 use OSM data uploaded to the region.",
+    "notice": "After clicking \"Create\" you will need to wait a few minutes for new OSM and GTFS feeds to be processed.",
+    "createDescription": "A Network Bundle requires both GTFS and OSM data. When creating a new Network Bundle you have the option of reusing data from previously created Network Bundles within this region. Note: Network Bundles created before April 15th, 2020 use OSM data uploaded to the region.",
```
ansoncfit, line 57:
```suggestion
    "createDescription": "A Network Bundle requires both GTFS and OSM data. When creating a new Network Bundle you have the option of reusing data from previously created Network Bundles within this region.",
```

I think we no longer need the note about OSM moving from region to bundle six months ago. If someone hasn't logged in since then, they probably don't remember the old setup anyways.

lib/components/create-bundle.tsx
```diff
@@ -151,11 +152,17 @@ export default function CreateBundle() {
         )}
 
         <Text>{message('bundle.createDescription')}</Text>
+
+        {!uploading && !error && (
+          <Alert status='info'>
+            <AlertIcon />
+            {message('bundle.notice')}
+          </Alert>
+        )}
```
ansoncfit, line 161: Maybe the `bundle.processing` message should move here? 

For consistency, I'd prefer the blue info box to be above or below the `bundle.createDescription` message, not switch -- but that's only a slight preference.

ansoncfit (commented):
Nice clarification, and I like seeing DocsLink in action.

Two suggestions are below.

lib/components/create-bundle.tsx
trevorgerhardt: Ah yea, I see that. Will put it below `createDescription`

ansoncfit (approved):

PR #1276, opened 2020-10-06
Clear destination point set id when selecting a regional analysis

### LINKED ISSUE #1273: Error switching between regional analysis results with different destination pointsets
**Describe the bug**
When viewing a completed regional analysis that has multiple available destination pointsets, if destinationPointSet `123` is selected via the UI, it is added to the URL. 

But this id is not cleared when navigating back to the list of regional analyses to view another one. 

So if the other has destination pointset `456` but not `123`, this error is displayed: `Error while communicating with server: Bad Request`. The console shows Destination gridId must be one of: `456`.

In addition to fixing the bug, we might want to consider displaying the more verbose error messages.

**To Reproduce**
View a completed regional analysis with multiple destination pointsets, and select one other than the first. Navigate back to the list of regional analyses, and choose one with a set of destination pointsets disjoint from the set of destination pointsets in the initial regional analysis.

trevorgerhardt:
It may not match up with a newly selected regional analysis. The comparison data is fine to keep, because that is a analysis/pointset pair.

Also, refactor setSearchParameter to be simpler, doesn't need to be overloaded.

Fixes #1273


ansoncfit (approved):
Main fix looks good. When selecting a "Weighted by" opportunity dataset, I would prefer not to show that on the map, but that can be a separate change.

PR #1279, opened 2020-10-06
Update recommended r5 version to v6

trevorgerhardt:


abyrd (approved):
Worker v6.0.0 is now released.

PR #1280, opened 2020-10-07
Fix adjust speed map ordering

trevorgerhardt:
Ensures the modified pattern is displayed above the original.


ansoncfit (commented):
Works when viewing the modification itself, but I'm not seeing the modified hops when displaying the modification on the map from the modification list.

trevorgerhardt:
> Works when viewing the modification itself, but I'm not seeing the modified hops when displaying the modification on the map from the modification list.

Fixed 👌

ansoncfit (approved):
🙏

PR #1283, opened 2020-10-07
Prevent flash of out of date content when saving an Editable

trevorgerhardt:
When saving a scenario name or modification name it's possible to see the old name flash for a second. This change waits for the editable component to finish the `onChange` method (which is usually a fetch) before reverting the editable component to the view state.


abyrd (approved):
I was able to reproduce the problem, and in testing this change does make the editing process more reassuring, since you don't see the old name come back momentarily.

PR #1288, opened 2020-10-07
Add inputs for older r5 versions

trevorgerhardt:
PR #1287 removed the ability to select a travel time percentile for regional analysis. This PR adds the ability to edit the cutoff and percentile for older versions of r5 in the create regional analysis screen.


abyrd (commented):
This looks like a good approach, with a different dialog box shown for older workers that expects a single cutoff and percentile. I have tested that this comes up for older workers. The multi-dimensional dialog for new workers validates its inputs, and shows a red outline and disables the create button if they are invalid. But the dialog for the older workers doesn't seem to apply any validation (though if I'm not mistaken the code does set up validation with testCutoff and testPercentile).

abyrd (approved):
Looks good 🔢

PR #1289, opened 2020-10-07
Wrap map components with panes to ensure correct layering

trevorgerhardt:
Reroute layers were sometimes showing the removed "red" path under the original path. This wraps the components in `Pane`s to ensure correct ordering.

A `<div position="relative">` is used to prevent the element the Pane's generate from flowing into the sidebar.

### DIFF
--- a/lib/components/analysis/create-regional.tsx
+++ b/lib/components/analysis/create-regional.tsx
@@ -53,8 +53,11 @@ const createTestArray = (min, max) => (sorted) =>
   sorted[0] >= min &&
   sorted[sorted.length - 1] <= max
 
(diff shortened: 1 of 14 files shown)


abyrd:
On the deployment of this PR, I'm getting NPEs on GraphQL requests whenever I view a modification, including a newly created reroute modification.

trevorgerhardt:
Is that from this PR or is that also happening on conveyal.dev ?

abyrd:
Not related to the PR, I just started typing here until I realized it was a backend issue. I'm moving the above comment to a backend issue.

abyrd (approved):

PR #1301, opened 2020-10-09
Cypress Test Refactor

### LINKED ISSUE #1192: Selector for opportunity datasets in regional analysis modal should include dataset source name
If someone imports two different shapefiles with "Jobs" attributes, for example, they can't be distinguished when selecting opportunity datasets for running a regional analysis.

### LINKED ISSUE #1315: Thrown lnglat errors prevent pages from loading
**Describe the bug**
A user appears to have made an Add Streets modification that goes over the anti-meridian. This throws an error when a selector attempts to determine the bounds of the modificaiton (here: https://github.com/conveyal/analysis-ui/blob/master/lib/selectors/modification-bounds.js#L14-L31).

**To Reproduce**

1. Create an `Add Streets` modification where the line goes over the anti-meridian.
2. Save. Try to open again.

**Expected behavior**
A few things:
1. The page should load. Even with the invalid data. So that the user can correct it.
2. We should prevent users from being able to create this kind of invalid data.

**Additional context**
lnglat throws for "invalid" coordinates. To prevent undefined behavior, all usage of that library should be checked or wrapped with a `catch` block.

trevorgerhardt:
### 1. Utilize TypeScript in `/cypress`
Code completion is huge for discovery. There are a ton of `cy.` commands and I've written a few dozen custom commands already. Cypress required it's own special TypeScript setup so this took a bit of finagling. 

In the future, I think we may convert to using an entirely separate `/test` directory that has it's own package.json so that `cypress` doesn't get installed by default and this configuration.

### 2. Refactor Test Setup
No longer needs to save the ids of the created entities. It merely looks for the name given. This means that the region name needs to be unique (which is useful anyway) and things like a project or modification have to have a unique name within their given entity.

We have a default region/bundle/project/opportunity dataset that's created automatically with the `nky` data in the fixtures folder.

Setup occurs inside of a `define()` but outside of a test. This allows for a cleaner initialization and sharing of resources. Decisions were heavily optimized for the development experience. Optimizations can be made later to make the entire test suite run faster.

 For example, a new regional analysis can be run with a specific analysis configuration but using the default data in less than 10 lines:
```js
import {getDefaultRegion} from './utils'
define('Fast walker test', () => {
  const region = getDefaultRegion()
  const regionalAnalysis = region.getRegionalAnalysis('Fast walkers', {settings: {walkSpeed: 2}})
  it('should have specific data related to fast walkers', () => {
    regionalAnalysis.navTo()
    // You are now viewing a completed regional analysis
  })
})
```
By default, the entity will persist across runs. It will just look for the created entity and grab the necessary data so that you can use it later (mainly for navigating to or deleting).

In continuous integration there will not be existing data so each run it will run through the creation process. Eventually, we will point this at staging and after an initial run the data will exist and we'll have our "Existing Region" tested.

The layout of the tests will surely change as we continue to add more. But this framework should help for quickly getting to the area that you want to test.

See `cypress/integration/example.ts` for a more extensive look at how to set up a test with custom data.

### 3. Related issues
- fixes #1192 
- fixes #1315


abyrd (approved):
Reviewed together in a work session. Looks like a very useful batch of changes.

PR #1370, opened 2020-11-24
Auto select modification feed

### LINKED ISSUE #1327: If only one feed is in bundle, it should be selected when creating new modifications
This would save users one click in the "Select Feed" dropdown.

trevorgerhardt:
Auto select the first feed of a bundle when creating a modification.

closes #1327


lib/components/modification/create.tsx
```diff
@@ -78,9 +79,9 @@ export default function CreateModification({
   async function create() {
     setIsCreating(true)
     const type = tabIndex === 0 ? transitTypeInput.value : streetTypeInput.value
-    const m = await dispatch(
+    const m: CL.IModification = await dispatch(
```
ansoncfit, line 82: Will this type enforcement be a problem if `ModificationTypes` has not been implemented for all modification types (see https://github.com/conveyal/analysis-ui/pull/1370/files#diff-19e42043c57ef6866bce171b4b6f4a9f8cfc4ac8ba3a48ea5b21d1b6985b4418R102 and above)?

trevorgerhardt: Short answer, it will not. 

Long answer: TypeScript is very helpful for compilation errors, autocompletion, and code hints but has no effect on the run time. It does aid in transpilation, with Babel.

In this specific instance, the result of  `await dispatch` is an "unsafe" `any` type. So we can cast it to anything we want.

In strict mode `any` is not allowed. Some day we will enable that...

ansoncfit (approved):
Looks good on staging

PR #1386, opened 2020-12-11
Add December changelog

trevorgerhardt:
Feel free to add/edit the text!


ansoncfit (approved):
Nice to have one changelog entry every two months in 2020!

For the next round of updates, we could consider pagination so the page doesn't get too long (and so old updates aren't as prominent).

PR #1392, opened 2020-12-21
Less obtrusive version warning

ansoncfit:
The warning that previous regional analyses used a different version of R5 took up a lot of vertical space. This PR replaces that alert with a warning icon that shows a Tip on hover:

image


lib/modules/r5-version/components/selector.tsx
```diff
@@ -80,9 +81,16 @@ export default function SelectR5Version({onChange, value, ...p}) {
...
-      <FormLabel htmlFor='select-r5-version'>
-        {message('r5Version.title')}
-      </FormLabel>
+      <Flex justify='space-between'>
+        <FormLabel htmlFor='select-r5-version'>
+          {message('r5Version.title')}
+        </FormLabel>
+        <div>
+          {lastUsedVersion && lastUsedVersion !== value && (
+            <WarningNote msg={message('r5Version.analysisVersionDifferent')} />
```
trevorgerhardt, line 90:
```suggestion
            <Tip label={message('r5Version.analysisVersionDifferent')}>
              <Box color='yellow.500'>
                <Icon icon={faQuestionCircle} />
              </Box>
            </Tip>
```

lib/modules/r5-version/components/selector.tsx
```diff
@@ -101,12 +109,6 @@ export default function SelectR5Version({onChange, value, ...p}) {
           {message('r5Version.latestReleaseVersionNotSelected')}
```
trevorgerhardt, line 109: Would we want this warning to still be in an <Alert /> below the select box?

trevorgerhardt (commented):
I'm not against this change, but it does hide the alert quite a bit more. Would we want both r5 version alerts to be hidden behind the tooltip? 

I don't think we need a custom warning component for this, added a suggestion to do it inline.

lib/modules/r5-version/components/selector.tsx
ansoncfit: Let's keep this warning as a more visible alert for now? If people are using an outdated version, we want to prompt them here to upgrade. But if they're using the updated version, they may be fine not re-running past analyses, so we make that message a bit less obtrusive here.

trevorgerhardt (approved):

Issue #1391, opened 2020-12-17
Freeform pointset changes


ansoncfit:
- [x] Remove DEV restriction on freeform pointset fields in opportunity dataset upload
- [x] Set `toLat` and `toLon` in analysis request when destination pinned
- [x] Do something with the `pathSummaries` returned in the JSON block

Related changes on the way at: https://github.com/conveyal/r5/pull/667
- [x] Add optional `originPointSetId` selector to analysis panel, filtering to freeform pointsets
- [x] If a regional analysis has `originPointSetId`, show CSV download button (see new endpoints added in https://github.com/conveyal/r5/pull/634/files)

PR #1404, opened 2020-12-23
Enable freeform point set upload

trevorgerhardt:
- [x] Remove `DEV` wrapper from point set checkboxes
- [x] Remove `(beta)` label from messages 

related to #1391


ansoncfit (changes requested):
The backend populates a default count value if the supplied `countField` is null, but not if it is blank: https://github.com/conveyal/r5/blob/b4a0756ad85ad0118783abb18665852e90e4ecdf/src/main/java/com/conveyal/r5/analyst/FreeFormPointSet.java#L117-L118

So we could not mark `countField` as required in the UI, if it's easy to signal null (by excluding it from the parameters sent?) instead of sending blank. Then users would only set it if they had a count column and wanted to override the default behavior (count is 1 per row). Sorry for not describing this correctly in our earlier discussion.

The same applies for the `idField`; but in most cases users will need to populate an idField anyways for post-processing, so it makes sense to keep it as required here.

Side note: I noticed that the fields marked `required` show a `webkit-validation-bubble-message` if someone tries to submit with a blank entry 👍

ansoncfit (approved):
Thanks for the updates 👍

PR #1405, opened 2020-12-23
Remove dependabot from update.always

trevorgerhardt:


ansoncfit (approved):
This will automatically merge dependabot updates that pass tests, via Kodiak.

We'll try it for a bit, but we probably won't keep this in place long-term given our planned transition from Kodiak to GH Automerge.

PR #1413, opened 2020-12-30
Add toLat/toLon to the request

trevorgerhardt:
In order to enable path results.

After running an analysis a circle marker appears under your mouse on the map that will show you a travel time distribution to that point. If you "lock" the marker in place by clicking on the map the currently displayed toLat/toLon will be added to the analysis request.

Addresses #1391


ansoncfit (approved):
Confirmed the coordinates are set on staging 👍

PR #1426, opened 2021-01-04
Wrap ckmeans usage in try/catch to prevent error

trevorgerhardt:
If the total data values are less than the breaks generated, an ugly uncaught exception will occur. This wraps that exception in a try/catch and shows an error message instead.

## Note
I couldn't reproduce the error locally. And therefore was unable to create a Cypress test for this. If someone knows how to reproduce, please let me know.

Addresses #1153

### DIFF
--- a/lib/actions/analysis/parse-times-data.ts
+++ b/lib/actions/analysis/parse-times-data.ts
@@ -6,9 +6,9 @@ const TIMES_GRID_TYPE = 'ACCESSGR'
 /**
  * Parse a grid header created by a GridResultWriter
  */
-export function parseGridHeader(ab: ArrayBuffer) {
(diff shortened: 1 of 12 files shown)


ansoncfit (approved):
Test case:

Set both cutoffs to 15 minutes [here](https://conveyal.dev/regions/5e566af61b2f767456cc8666/regional?analysisId=6019aa7c35caca36578d8fce&cutoff=15&comparisonAnalysisId=6019aa6835caca36578d8fcc&comparisonCutoff=30&comparisonPercentile=50&comparisonDestinationPointSetId=6019a93535caca36578d8fc9).

That example shows an arcane error message on dev, and works on this branch with message "Data not suitable for generating a color scale." ✔

PR #1428, opened 2021-01-09
Update R5 version warning for street modifications

ansoncfit:
Specify version v5.9.0 _or higher_, and make the alert an orange warning instead of a red error


trevorgerhardt (approved):
👍

PR #1436, opened 2021-01-18
Bugfix: use correct index when dragging stops

### LINKED ISSUE #1432: Problems editing alignments downstream of control points
**Describe the bug**
When editing alignments, changes to a given stop point (moving, converting to/from control point, deleting) are incorrectly applied to the nth preceding stop/control point, where n is the number of control points in the alignment upstream of the stop point being changed.

Two examples, when editing the alignment of an add-trip modification:

**Example 1**
1. Draw 2 stops
2. Convert the 2nd stop to a control point
3. Add a 3rd stop (extending from end)
4. Drag the 3rd stop to move it. The control point will be moved instead.

**Example 2**
1. Draw 3 stops
2. Convert the 3rd stop to a control point
3. Convert the 2nd stop to a control point
4. Add a 4th stop (extending from end)
5. Move the newly added stop -- The first control point will be moved instead.
6. Try converting the end stop to a control point. Nothing will happen.

trevorgerhardt:
Incorrect `stopIndex` was being passed to `onDragStop`. This adds a test with the examples from #1432 and fixes the problem by using the correct stop index. 

Other changes include using the specific array index for each point type for its key/title.

fixes #1432

### DIFF
--- a/cypress/integration/modifications/add-trip-pattern.ts
+++ b/cypress/integration/modifications/add-trip-pattern.ts
@@ -130,13 +130,13 @@ describe('Add Trip Pattern', () => {
       const newCoord: L.LatLngTuple = [39.08, -84.5]
       const tempCoord: L.LatLngTuple = [39.09, -84.5]
 
-      cy.dragMarker('Stop 1', lastCoord, newCoord)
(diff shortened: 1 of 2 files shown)


ansoncfit (approved):
Thanks!

PR #1450, opened 2021-02-08
OD upload status

ansoncfit:
In upload status messages, "features" should be "layers"


trevorgerhardt (approved):

trevorgerhardt:
@ansoncfit looks good but there appears to be a Cypress test that waits to see `feature`. Can you update that test?

PR #1451, opened 2021-02-08
Add ZenDesk chat widget

trevorgerhardt:
Add's the snippet from ZenDesk for the chat/support widget. If no "agents" are online (which will be the usual case), it will allow users to directly submit support requests from inside the application.

This PR also identifies all logged in users with the widget so that user's will not have to input their email addresses for us to know who submitted the request/chat.

### DIFF
--- a/lib/components/sidebar.tsx
+++ b/lib/components/sidebar.tsx
@@ -5,9 +5,9 @@ import {
   faCubes,
   faDatabase,
   faGlobe,
+  faInfoCircle,
   faLayerGroup,
   faPencilAlt,
-  faQuestionCircle,
   faServer,
   faSignOutAlt,
   faTh,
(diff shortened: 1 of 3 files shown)


ansoncfit (commented):
Works well on staging.

Duplicating the question mark icon in the lower left (docs link) and lower right (widget) feels a bit strange. Assuming it's tricky to customize the stock widget, could we update the docs icon to 📖?

Is it easy to minimize/hide the Help button? Not crucial, but that might be useful for people who are running workshops.

trevorgerhardt:
> Works well on staging.
> 
> Duplicating the question mark icon in the lower left (docs link) and lower right (widget) feels a bit strange. Assuming it's tricky to customize the stock widget, could we update the docs icon to 📖?
> 
> Is it easy to minimize/hide the Help button? Not crucial, but that might be useful for people who are running workshops.

Yea, we can update the docs icon. I'll see how the book looks, could also try an `(i)`?

It is possible to hide the button from the console...I think. See docs here: https://developer.zendesk.com/embeddables/docs/widget/core#hide

trevorgerhardt:
Screen Shot 2021-02-09 at 16 01 50

trevorgerhardt:
Screen Shot 2021-02-09 at 16 03 07

trevorgerhardt:
I think the `i` works a bit better there. But let me know what you think @ansoncfit

ansoncfit (approved):
Thanks for switching the logos

PR #1454, opened 2021-02-10
Upgrade to react v17 to utilize latest nextjs features

trevorgerhardt:
Next.js is showing warnings that certain features (like Fast Refresh in dev mode) no longer work with older versions of React. Also, v16 -> v17 seems to be a safe upgrade for us here.


ansoncfit (approved):
👍

PR #1455, opened 2021-02-10
Fix destination lock

### LINKED ISSUE #1431: Destination no longer locked for travel time distribution legend
After https://github.com/conveyal/analysis-ui/pull/1413, clicking to lock the destination in a single-point analysis does lock the destination on the map, but the travel time destination continues to follow the cursor after results are refreshed.

trevorgerhardt:
closes #1431


ansoncfit (approved):
Tested on staging, looks good

PR #1456, opened 2021-02-12
Show errors in create modal

trevorgerhardt:
If an error occurs while creating a regional analysis, log it to the console and show it in an `Alert` box. The dispatched fetch was causing errors to be handled in a variety of ways, so I moved it to use a `safeFetch` to handle errors in the same way. (`safeFetch` should never throw an error)

### DIFF
--- a/cypress/integration/clear-data.ts
+++ b/cypress/integration/clear-data.ts
@@ -1,6 +1,6 @@
 import {clearTemporaryData} from './utils'
 
-describe.skip('Clear all test data', () => {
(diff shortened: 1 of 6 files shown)


ansoncfit (changes requested):
When I tried this on staging (using Test Region - Boston), this did not surface a 400 returned by the server after trying to create a regional analysis with destinations with different extents.

image

image

ansoncfit (approved):
Just tested on staging 👍

PR #1457, opened 2021-02-12
Proxy backend to avoid CORS configuration

mattwigway:
## Description

This proxies the backend server through /api/backend when in dev mode using `http-proxy-middleware`, so that CORS headers are not needed locally. Adds an env var NEXT_PUBLIC_BACKEND_PROXY_URL=http://localhost:7070 to proxy in dev mode, and changes NEXT_PUBLIC_API_URL to /api/backend. No changes to production.

Fixes # (issue)

## How to test

1. Start up R5 (I'll add a pull request that remove the CORS headers when running locally)
2. Use analysis-ui dev, confirm that it works, look in developer panel to see that requests are being proxied to the backend with no CORS headers.


mattwigway:
Goes with conveyal/r5#683

abyrd:
See discussion on conveyal/r5#683 : I think we should just make the CORS headers configurable on the backend and supply a default that will work with analysis-ui out of the box.

trevorgerhardt:
With https://github.com/conveyal/r5/pull/683 this PR should be unnecessary.

PR #1458, opened 2021-02-12
Select origin point set

### LINKED ISSUE #1168: Don't redirect to regional analysis page after creating one
For now, use chakra 🍞 with 🔗 instead.

Longer-term, make common progress reporting available.

trevorgerhardt:
- [x] Change to selector with first option as `Rectangular Grid`. 
- [x] Show total number of origin points from grid or point set

Screen Shot 2021-02-12 at 17 40 21

Defaults to "Full Region". When "Full Region" is selected, selector is hidden.

Thoughts on style and text @ansoncfit? 

addresses #1391 
closes #1168


lib/components/analysis/create-regional.tsx
```diff
@@ -96,17 +102,49 @@ export default function CreateRegional({
   )
 }
 
+/**
+ * Show this toast when a modification has been copied.
```
abyrd, line 106: This comment does not need to describe the role of the toast below.

abyrd:
In our terminology, a grid covering the whole region and freeform custom points are both "point sets". The text `Origins from [] Full Region [] Point Set` implies that the full region is not a point set and vice versa, which could get confusing. In addition, the grid is not always for the whole region - its extent can be set.

So the real choice is between a rectangular grid and non-gridded custom points. An alternative mechanism could be a simple checkbox for "gridded origins" which when unchecked reveals the list of custom origins. But that doesn't really explain to the user what they are selecting when the box in unchecked.

So I'd suggest: `Origin points [] on rectangular grid [] from custom upload`. And maybe the whole section should only appear when the region has any freeform pointsets uploaded, as most users won't have any at this point. Or if no freeform pointsets are available, it could show a short explanation and link to the CSV upload page.

ansoncfit:
`Origin points [] on rectangular grid [] from custom upload` 👍, maybe with a small box with a dashed blue line after rectangular grid to remind users that this is the grid defined by the (possibly custom bounds) show on the map

Next step: show number of origins in this box

lib/utils/calculate-grid-points.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export default function calculateGridPoints(
+  bounds: CL.Bounds,
+  zoom = PROJECTION_ZOOM_LEVEL
+): number {
+  const topLeft = lonlat.toPixel([bounds.west, bounds.north], zoom)
+  const bottomRight = lonlat.toPixel([bounds.east, bounds.south], zoom)
+  const width = topLeft.x - bottomRight.x
+  const height = topLeft.y - bottomRight.y
+  return width * height
```
ansoncfit, line 13: Round this to avoid:

image
?

lib/utils/calculate-grid-points.ts
```diff
@@ -0,0 +1,14 @@
+import lonlat from '@conveyal/lonlat'
+
+import {PROJECTION_ZOOM_LEVEL} from 'lib/constants'
+
+export default function calculateGridPoints(
+  bounds: CL.Bounds,
+  zoom = PROJECTION_ZOOM_LEVEL
```
ansoncfit, line 7: Use the request zoom instead of a constant?

lib/components/analysis/create-regional.tsx
```diff
@@ -159,36 +197,47 @@ function CreateModal({onClose, profileRequest, projectId, variantIndex}) {
...
       setIsCreating(false)
-      setError(response.data.message)
-    } else {
-      const {as, href} = routeTo('regionalAnalyses', {regionId})
-      router.push(href, as)
+      if (response.data instanceof Error) {
+        setError(response.data.message)
+      } else {
+        setError(response.problem)
+      }
```
ansoncfit, line 240: The error below displays "CLIENT_ERROR" in the UI, but we probably want the message:

image

lib/utils/calculate-grid-points.ts
trevorgerhardt: The constant is just the default. A profile request right now doesn't have a zoom level set by default on the client side, so unless we set that it will be `undefined` by default and use that `PROJECTION_ZOOM_LEVEL` constant.

trevorgerhardt: I also meant to say I do pass the request zoom as a parameter: https://github.com/conveyal/analysis-ui/pull/1458/files#diff-4bb0bd244bcd85f470907a24692a3c5931814d450fb9050a263f39554e8d29f4R159

trevorgerhardt: I also meant to say I do pass the request zoom as a parameter: https://github.com/conveyal/analysis-ui/pull/1458/files#diff-4bb0bd244bcd85f470907a24692a3c5931814d450fb9050a263f39554e8d29f4R159

trevorgerhardt:
#### 📝Changed routes:


| [**/regions/\[regionId\]/edit**](https://analysis-myhr84rrl.vercel.app/regions/[regionId]/edit) |
|:-:|
|<a href="https://analysis-myhr84rrl.vercel.app/regions/[regionId]/edit">Screenshot of /regions/[regionId]/edit</a><br /><sup><a href="https://images.weserv.nl?url=https://s.microlink.io/1xxsyImRQ010SJpWSmA1eeR8oBs_.png">(view full size)</a> |







Commit 5863d1c59b0fe9967f4822f144564cbf4737fb88 (https://analysis-myhr84rrl.vercel.app).

lib/components/analysis/create-regional.tsx
trevorgerhardt: Updated the structure of the error responses to be more consistent 👍

lib/utils/calculate-grid-points.ts
trevorgerhardt: Rounded 👍

ansoncfit (approved):
Thanks for the updates

PR #1463, opened 2021-02-17
Change Opportunity -> Spatial, display only

trevorgerhardt:
Change display values from `Opportunity` to `Spatial` where the more generic term should be used. 

Note: does **not** rename any variable names.


ansoncfit (approved):
Looks good

PR #1465, opened 2021-02-18
Log paths for debugging

trevorgerhardt:
Logs entire parsed response, which contains `pathSummaries` when enabled.

Addresses #1391


ansoncfit (approved):
🙏

PR #1469, opened 2021-02-23
Spatial dataset labels

### LINKED ISSUE #988: Clarify opportunity data management options
I've found the terminology around managing opportunity datasets to be fairly misleading/confusing. 

Screenshot_2020-01-23_19-48-47

"Dataset" is used to mean two different things here and the distinction between a "dataset" and a "dataset source" is not clear. There are two basic objects in play here: a spatial dataset and the variables associated with it. You can edit the names of the variables and delete them. And you can delete the whole spatial dataset along with all associated variables. 

I would suggest that the terminology on these buttons be changed as follows:
* _Upload a new dataset_ -> Upload a new spatial dataset
* _Fetch LODES (2015)_ -> (no change)
* _Edit dataset name_ -> Edit selected variable name
* _Delete this dataset_ -> Delete selected variable
* _Delete entire dataset source_ -> Delete entire spatial dataset

Though that last one still wouldn't be clear since it's not totally obvious which dataset is selected. Only the variable name is shown and the popup are-you-sure message doesn't give the name either. The easiest thing would probably be to add the name of the spatial dataset to the pop-up.

The terminology on the Analysis tab is fairly intuitive already IMO since both the spatial dataset name and the variable name are kept together.

ansoncfit:
Follow-up to #1463, should fix #988

### DIFF
--- a/cypress/integration/2-opportunities.ts
+++ b/cypress/integration/2-opportunities.ts
@@ -141,7 +141,7 @@ describe('Opportunity Datasets', function () {
       .clear()
       .type(opportunity.longitudeField)
 
-    cy.findByLabelText(/Enable freeform/).check({force: true})
+    cy.findByLabelText(/Enable freeform (non-grid) points/).check({force: true})
     cy.findByLabelText(/ID field/).type('sport')
     cy.findByLabelText(/Opportunity count field/).type('count')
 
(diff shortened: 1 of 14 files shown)


trevorgerhardt (approved):
LGTM.

PR #1471, opened 2021-02-23
Filter out free form datasets from destination selector

trevorgerhardt:
- Adds a filter prop to the OD Selector.
- Filters `format != 'FREEFORM'`


ansoncfit (approved):
Looks good, thanks

Issue #1118, opened 2020-06-17
Single digit hours fail validation, but analysis is still allowed


abyrd:
## Observed Behavior
When doing a single point analysis, when a start time is entered that has a one-digit hour but is not prefixed with a zero, it fails validation. This is highlighted with a red outline. However, it is still possible to start an analysis which will use whatever time was previously accepted by the validation.

## Expected behavior
1. One-digit hours should be accepted as valid
2. When inputs are invalid, analysis should probably be blocked. The user may believe that results displayed reflect the visible inputs that failed validation.

trevorgerhardt:
Currently, each input handles its own validation. We encapsulate the inputs in component groups to help organize the analysis settings. In order to determine if the entire form is valid we would need to connect the form inputs with a top level. Therefore this change is not straightforward. 

I began work on implementing a "FormContext" that would help handle this. But it could turn into quite a large change set so I moved this issue back into the January release for now. (See https://github.com/conveyal/analysis-ui/pull/1312)

abyrd:
In the mean time can we change validation to allow single digit hours? That would greatly reduce the incidence of the problem.

PR #1480, opened 2021-02-26
Limit transfers to 1 character (single digit)

trevorgerhardt:
The `maxLength` attribute limits the total number of characters allowed to be typed into the input. A min/max number limit only helps with the form validation.

### DIFF
--- a/lib/components/analysis/__tests__/__snapshots__/profile-request-editor.js.snap
+++ b/lib/components/analysis/__tests__/__snapshots__/profile-request-editor.js.snap
@@ -66,7 +66,6 @@ exports[`Components > Analysis > Profile Request Editor should render correctly 
       maxLength={1}
       onChange={[Function]}
       size="md"
-      type="number"
       value={3}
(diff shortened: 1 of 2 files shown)


ansoncfit (commented):
In Chrome (88.0.4324.190), I can type in a multi-digit transfer limit (doesn't align with my interpretation of "allowed to be typed" in the comment on this PR) and the box turns red.

Eventually address with https://github.com/conveyal/analysis-ui/issues/1118?

trevorgerhardt:
> In Chrome (88.0.4324.190), I can type in a multi-digit transfer limit (doesn't align with my interpretation of "allowed to be typed" in the comment on this PR) and the box turns red.

Browser quirks? If we remove `type='number'` (which we don't actually need here) it respects the `maxLength` attribute. 

Please try again.

ansoncfit (approved):
Works well 👍

It is now possible to enter non-numeric characters in this field, but I don't think we should worry about that.

Issue #866, opened 2019-04-04
Multiple collapsible profile requests


ansoncfit:
Currently, two isochrones can be displayed with the only difference being the scenario vs. comparison scenario. It would be helpful to be able to superimpose isochrones reflecting differences in other profile request parameters (e.g. access modes, departure time window, reliability percentile, etc.)

ansoncfit:
Idea from @abyrd related to how display results chart in a unified way once settings for each request have been split out:
> Display destination travel time box-and-whisker bar below the cumulative opportunities plot, reusing its time axis (instead of in a popup). With reverse-geocoded name?

trevorgerhardt:
Closed in #1097

PR #1481, opened 2021-02-26
Add "Adjust Dwell Time" scratch tests

trevorgerhardt:
- Add a text input to adjust the `Time cutoff` in the Single Point Analysis page
- Fix Mode selector spacing on Analysis page that changed in the Chakra v1 migration


lib/components/analysis/results-sliders.tsx
```diff
@@ -57,6 +57,18 @@ export function CutoffSlider({isDisabled, ...p}) {
...
+        onChange={(e) => {
+          const v = parseInt(e.currentTarget.value)
+          if (!isNaN(v)) cutoffSlider.onChange(parseInt(e.currentTarget.value))
+        }}
+        style={{
+          display: 'block',
+          height: 0
+        }}
+        value={cutoffSlider.value}
+      />
```
ansoncfit, line 71: Relative to the current dev branch, this change seems to add a small stutter in the cutoff update when dragging the slider (especially when a destination opportunity layer is loaded).

Is there an easy way to memoize or disable this input except in testing?

trevorgerhardt: Hmmm, yes I see that.

Assuming I can make it smooth, should we make this text input available next to the slider anyway? Something like:

Screen Shot 2021-02-26 at 14 47 46

ansoncfit: The x-axis of the cumulative opportunity plot used to align perfectly with the travel time slider. Could be worth revisiting here?

trevorgerhardt: We'd have to put the label and input on a separate line then. If anything I'd prefer to disassociate it a bit more (make it smaller?) so that the expectation of alignment isn't there.

trevorgerhardt: Like this:
Screen Shot 2021-03-03 at 23 10 30
thoughts?

ansoncfit: We previously discussed the idea of switching the travel time percentile to a vertical slider, to the left of the cumulative opportunities plot (at https://github.com/conveyal/analysis-ui/issues/866#issuecomment-579292566). I recognize that's a big change, but I think a vertical slider with 5 at the top and 95 at the bottom, juxtaposed with the vertically oriented box-and-whiskers plot of the accessibility distribution, would make this parameter easier to understand.

The cumulative opportunities plot would then be narrower, with its left edge translated to the right, making room for the time cutoff box to the left of the slider.

Again, happy to revisit in a separate PR, as this is a bit of a tangent from the tests that are the focus of this one.

trevorgerhardt: Yea, I'll sketch it out to see how it will look and we can see if we want to go in that direction. I don't think we'd want to change this twice.

trevorgerhardt: So, first thought is that `Travel time percentile` is relevant whenever there are results, even when an opportunity layer is not selected, but the plot only shows up when an opportunity layer is selected.

trevorgerhardt: I'm not too enthused about the look either.

Screen Shot 2021-03-04 at 08 15 56

ansoncfit (approved):
Thanks for addressing the slider smoothness and alignment.

My only remaining comment is the destination selector should be wider, to avoid wrapping the options:
image

But let's merge the new tests now, and I'll open a subsequent PR for spacing etc.

Issue #1313, opened 2020-10-13
Add max size limits for each file upload


trevorgerhardt:
We recently had a user attempt to upload an OSM file > 400mb. We should add size limits to each file upload.

File size limits for each TBD:
- GTFS (?)
- OSM (?)
- Opportunities (?)
- Importing shapefiles to project (?)
- Aggregation Areas (?)

trevorgerhardt:
@abyrd and @ansoncfit thought on default max sizes that we should use for each?

trevorgerhardt:
I'd imagine 1GB for anything would be too large. What's the largest GTFS feed we've seen?

abyrd:
In our deployment, our nginx configuration has a `client_max_body_size` of 500MB, so anything larger than that will already fail. I agree it would probably be good to validate those sizes client side before the upload to avoid confusion / disappointment when they fail. The Netherlands feed is weighing in at 250-350MB these days. Most other national scale GTFS I've encountered is smaller. 

For OSM, a Netherlands extract is over 1GB and Norway is over 700MB. However most of that is buildings, water etc. The Netherlands filtered down to just routable ways is only a bit over 100MB, which also means it's much faster to upload and process. I'd really like to encourage people to make this 10x improvement but I guess we can't force that.

Shapefiles can also be quite huge - similar observations there. Properly optimized ones are often not so bad, but we can't ask all users to spend time optimizing their file sizes. 

My sense is that the 500MB size limit in nginx is working well.

Maybe when selecting files, if size information is available and the file is seen to be large, we can display a message with a few tips on shrinking PBF files before the user uploads it.

trevorgerhardt:
Closed in #1483

PR #1483, opened 2021-02-26
Add size limit to OSM and GTFS upload

### LINKED ISSUE #1478: Warn on upload of large files
Our reverse proxy has body size limit of 500MB, which limits GTFS and OSM uploads. Even for large regions, 500MB is a lot of OSM, which is usually mostly building outlines, park benches, and other things not needed for routing, which all get loaded and slow down processing. So it probably does make sense to impose a limit.

However when you upload a larger file, it just stalls creating the bundle and doesn't show any error. We should display a message about the max file size, or even check the file size client side and refuse to upload it if it's too big, with a link to the section of the docs on filtering OSM data.

trevorgerhardt:
- [x] Add limits to all other file uploads.

Add a 500MB limit to individual OSM and GTFS upload fields and a 1GB total limit. Show an error message when a user selects a file or files that exceed those limits. 

closes #1478 and #1313

## How to test

1. Select an OSM file that's more than 500MB and ensure that the error message shows and you cannot click the upload button.
2. Repeat for the GTFS field.
3. Select multiple files that are smaller than 500MB but exceed 1GB and ensure the error is shown.


abyrd (approved):
Seems to work well in testing. As expected it won't let me upload my planet.osm.pbf. MacOS confused me momentarily - the dialog box shows file sizes with the metric mega prefix, so `australia.osm.pbf` is 519.7MB but did not trigger the warning (it's actually 496MiB using binary mega=2^10). The toast text "Invalid File Selected / Each file has a maximum limit..." sounds a little odd to me. I'd suggest something like "Selected file too large. Each file is limited to..." but that's minor.

PR #1484, opened 2021-02-26
Return updated document in db PUTs

### LINKED ISSUE #1472: Presets modified name does not change until page reload

trevorgerhardt:
The original document is returned by default, this change adds the `returnOriginal: false` option to db PUTs in the JavaScript API. 

Fixes #1472


ansoncfit (approved):
Tested on staging 👍

PR #1485, opened 2021-02-26
Cache Cypress binary, skip initial installation

trevorgerhardt:
Recommendations taken from https://github.com/bahmutov/cypress-gh-action-split-install. This should fix the binary installation error we've been seeing in Github Actions.


ansoncfit (approved):
👍👍

PR #1492, opened 2021-03-01
Add API Status bar

trevorgerhardt:
Ping the `/version` endpoint of the configured `API_URL` and show an error bar if the API cannot be reached. Show a different error message if the API cannot be reached and the user is not connected to the internet. (If the API can be reached and the user is not connected that probably indicates the user is developing locally).

Checks the API on a 30 second interval and stops checking if the user is not active (See https://swr.vercel.app/docs/revalidation for how that works). 

**Please give feedback on the text displayed and display timing logic.**

This will show after 10 seconds if the API has not responded:

Screen Shot 2021-03-01 at 16 24 10

This will show if the user does not have an active internet connection (according to the [browser's web api](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine)) and the API responded with an error or response is still processing as it should be almost instantaneous locally:

Screen Shot 2021-03-01 at 16 23 28

This will show if the request returns with an error (most likely that will be a 500 error/timeout because the server is down):

Screen Shot 2021-03-01 at 16 23 19

Note: the new `Help` chat box should still show if they are online but our server is down.


abyrd:
See draft `/api/activity` endpoint at conveyal/r5#707. It currently reports OSM/GTFS upload progress. I will be cleaning up the task tracking and expanding this to report any OSM/GTFS errors encountered.

yarn.lock
```diff
@@ -4745,7 +4745,7 @@ depd@~1.1.2:
   resolved "https://registry.yarnpkg.com/depd/-/depd-1.1.2.tgz#9bcd52e14c097763e749b274c4346ed2e560b5a9"
   integrity sha1-m81S4UwJd2PnSbJ0xDRu0uVgtak=
 
-dequal@2.0.2:
+dequal@2.0.2, dequal@^2.0.2:
```
ansoncfit, line 4748: repeated?

lib/components/api-status-bar.tsx
```diff
@@ -0,0 +1,51 @@
...
+import BannerAlert from './banner-alert'
+
+const isValidatingTimeout = 10_000
+const unusableMessage =
+  'Application will be unusable until connection can be restablished.'
+
+const IsValidating = () => (
+  <BannerAlert status='warning'>
+    <AlertTitle>Establishing connection...</AlertTitle>
+  </BannerAlert>
```
ansoncfit, line 15: This might make the user feel like something is wrong, but it's an expected loading step.

Hide the warning icon, follow up with suggestion "Please wait," or add animation (e.g. https://github.com/szchenghuang/react-animated-ellipsis)?

lib/components/api-status-bar.tsx
```diff
@@ -0,0 +1,51 @@
...
+const NoAPI = () => (
+  <BannerAlert status='error' variant='solid'>
+    <AlertTitle>API server cannot be reached. {unusableMessage}</AlertTitle>
+  </BannerAlert>
+)
+
+const NotOnline = () => (
+  <BannerAlert status='error' variant='solid'>
+    <AlertTitle>
+      You are not connected to the internet. {unusableMessage}
```
ansoncfit, line 27:
```suggestion
      {unusableMessage} Check your internet connection.
```

lib/components/api-status-bar.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+const IsValidating = () => (
+  <BannerAlert status='warning'>
+    <AlertTitle>Establishing connection...</AlertTitle>
+  </BannerAlert>
+)
+
+const NoAPI = () => (
+  <BannerAlert status='error' variant='solid'>
+    <AlertTitle>API server cannot be reached. {unusableMessage}</AlertTitle>
```
ansoncfit, line 20:
```suggestion
    <AlertTitle>{unusableMessage} Contact your support team if this message persists.</AlertTitle>
```

lib/components/api-status-bar.tsx
```diff
@@ -0,0 +1,51 @@
+import {AlertTitle} from '@chakra-ui/react'
+import useStatus from 'lib/hooks/use-status'
+import useIsOnline from 'lib/hooks/use-is-online'
+import {useEffect, useState} from 'react'
+
+import BannerAlert from './banner-alert'
+
+const isValidatingTimeout = 10_000
+const unusableMessage =
+  'Application will be unusable until connection can be restablished.'
```
ansoncfit, line 10:
```suggestion
  'Lost connection to Conveyal servers.'
```

ansoncfit (commented):
Overall approach and suggested timing/logic look good to me.

I added a few suggested wording changes to the alerts, trying to include consistent and concise status message and suggested action.

yarn.lock
trevorgerhardt: It's an autogenerated file, but I believe that means there were two separate dependencies referenced for `dequal`, one that was `2.0.2` and one for `^2.0.2` and that the listed package below that meets the requirements for both.

lib/components/api-status-bar.tsx
trevorgerhardt: True, although this shouldn't show up unless the connection is really bad or the server is down. 
- I'll see how the compass spinner we use in the rest of the application looks.
- Also, can switch it to an `status='info'` style. 
- Does the full text `Establishing connection, please wait.` make sense?

trevorgerhardt: image

abyrd (approved):
We decided to merge this and adjust the logic if we see any spurious connection warnings over the coming weeks. I reverted the commit a231725 to switch back to /api/activity

abyrd:
Hmm Cypress tests failing since I changed to the /api/activity endpoint. dev should now have that endpoint.

trevorgerhardt:
> Hmm Cypress tests failing since I changed to the /api/activity endpoint. The R5 dev branch should now have that endpoint, and it seems to cooperate with the ui in local testing.

The build failed in R5: https://github.com/conveyal/r5/runs/2036376873, re-running

PR #1502, opened 2021-03-04
Refactor _app for a reduced initial payload

trevorgerhardt:
Recently while running `next build` I was seeing initial shared payloads over 200kb. A lot of our pages will require that or even double, but the initial load can and should be much smaller with more client only code available to lazy load later. This refactor takes a stab at reducing the shared initial payload getting it down to under 130kb. 

- Move error handlers into a single component.
- Use `next/dynamic` to lazy load error modals and toasts that are meant to catch client side application errors.
- Use `react-error-boundary` around the whole application.
- Move SWR and Chakra into `/configs`


ansoncfit (commented):
Testing with a running staging server, I'm seeing the lost connection status warning

trevorgerhardt:
> Testing with a running staging server, I'm seeing the lost connection status warning

Hmmm, yea, that's showing for staging also right now. Was the new staging server deployed with a version that contains the `/activity` API endpoint?

ansoncfit (approved):
Works with updated staging server

PR #1503, opened 2021-03-05
Confirm popover

### LINKED ISSUE #1486: Switch to popover for confirmations instead of a modal
If you click on "Delete X", a modal appears in the middle of the screen asking you to confirm "Deleting X". Instead, we could show it positioned right next to the element that initiated it using a [Popover](https://chakra-ui.com/docs/overlay/popover).

trevorgerhardt:
Use a Popover instead of a Dialog for ConfirmButton which allows the action confirmation to popup right next to the place it was triggered.

Screen Shot 2021-03-05 at 10 37 49

closes #1486

### DIFF
--- a/cypress/integration/0-regions.ts
+++ b/cypress/integration/0-regions.ts
@@ -61,7 +61,7 @@ function testInvalidCoordinates() {
 function deleteThisRegion() {
   // Delete region
   cy.findButton(/Delete this region/).click()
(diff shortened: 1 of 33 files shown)


ansoncfit (approved):
On staging, tested the new confirmation for deleting
* Region
* Bundle
* Project
* Preset
* Regional analysis

PR #1508, opened 2021-03-08
Refactor geocoder to work without react-select

trevorgerhardt:
Additionally, this allows us to put the geocoder manually onto the page to more directly control what occurs on a geocoder select. For example, on the Analysis Page we most likely want the marker location to be updated.

Screen Shot 2021-03-08 at 16 16 42


ansoncfit (approved):
Nice

PR #1512, opened 2021-03-16
Move CSS into individual components

trevorgerhardt:
This should make the initial page load faster. Removes the global CSS imports from `_document` and moves them into the components that require them. (Made possible by recent updates to Next.js)

- Removes now unused FontAwesome CSS
- The home page will no longer download leaflet or mapbox-gl.css
- Creates a reusable Marker component to set the image files correctly
- Creates a reusable leaflet draw component to ensure the CSS for that gets required (at most once)


ansoncfit (changes requested):
With fontawesome gone, `className='fa-spin'` should be updated in a couple places.

trevorgerhardt:
> With fontawesome gone, `className='fa-spin'` should be updated in a couple places.

Ah yes, forgot we were still using that.

Btw, we're still using FontAwesome icons, but now importing each icon directly as SVG through `react-icons`.

trevorgerhardt:
> With fontawesome gone, `className='fa-spin'` should be updated in a couple places.

Fixed 👍

ansoncfit (approved):

PR #1520, opened 2021-03-23
Adjust sliders, line up elements

trevorgerhardt:
- Add symbols
- Memoize components
- Refactor structure so that components don't re-render if unnecessary


abyrd (approved):
In Safari testing, moving the slider still tends to select the text in the cumulative plot legend or labels below the slider, but doesn't select the text on the slider handle itself, and doesn't hide the x-axis labels, which were the two most visible things.

PR #1522, opened 2021-04-09
Clock drift compensation and property name changes

abyrd:
As discussed recently, we have no guarantee that the server's clock is in sync with the client machine's clock. Especially since we're tracking tasks that may take significantly less than a minute, using absolute times reported by the server risks displaying durations that are negative or start well above zero. 

This PR derives a startTime field known to be accurate on the client machine from relative durations reported on the API. I may be breaking some conventions in the way I did this, feel free to push additional changes. This just serves as an example of the idea.

This PR is relative to`status-reporting` but branches off before the recent merge from `dev`, which seems to have broken a filename selector.

### DIFF
--- a/lib/components/task-list.tsx
+++ b/lib/components/task-list.tsx
@@ -6,6 +6,7 @@ import {
   Stack,
   StackDivider
 } from '@chakra-ui/react'
+import dateSubtract from 'date-fns/sub'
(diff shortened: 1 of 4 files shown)


abyrd:
Should work with r5 branch `activity-api` commit ecb7a10.

PR #1528, opened 2021-04-27
Use new strategy to help prevent font flash

trevorgerhardt:
- Use a local version of the font for offline usage.
- Use a CDN backed, immutable version on production with a very long lifespan that conveyal.com and the docs also use.

Related: https://github.com/conveyal/docs/pull/49 & https://github.com/conveyal/conveyal.com/pull/98


ansoncfit (approved):
👍

PR #1529, opened 2021-04-29
UI Cleanup

### LINKED ISSUE #1527: Miscellaneous pre-release UI cleanup
* After navigating to region/project, basemap is centered on previously used analysis area (not necessarily current region)
* Margins around list of stop names in Remove Stops modification:
image
* Analysis date truncated:
image

trevorgerhardt:
- Add left padding back to list of stops
- Fix: redirect map correctly to a region's center on region change.
- Reduce input padding to prevent calendar icon from hiding text where `input type='date'`. (Also, remove duplicate code by reusing the `InputWithUnits` component where appropriate).

fixes #1527


ansoncfit:
Additional observations
- Show/hide modification visibility icons are reversed from our previous convention
- Initial region list is horizontally scrollable

ansoncfit:
On my screen, the date still needs a bit more padding. Also, is the Fetch Results button intended to be larger than the Regional Analysis + buttons now?

image

trevorgerhardt:
> On my screen, the date still needs a bit more padding. Also, is the Fetch Results button intended to be larger than the Regional Analysis + buttons now?

- Reduced the padding a bit more, that should be enough but let me know if it isn't!
- I increased the size after hearing about the person who didn't know what to click. Not sure if that will help or if it's necessary. I can reduce if you'd like, I'm not wedded to it.

ansoncfit:
Padding looks good.

I think I slightly prefer the previous Fetch results button size -- I'll work on separate UI/docs changes to make it clearer when to press it.

ansoncfit:
The isochrone slider feels a bit laggy again.

Possibly related: when quickly dragging the slider left, and overshooting so the cursor ends up over the sidebar, the slider value drops to 0 (as expected) but the vertical slice and accessibility results do not catch up:

image

ansoncfit (approved):
Approving (including my changes in #1531, accidentally merged without review).

PR #1535, opened 2021-05-04
Cutoff Slider Updates

trevorgerhardt:
- Allow setting the cutoff to 0 (fixes an issue that prevented the graph from updating)
- Move props to prevent needless updates when the cutoff changes.


ansoncfit (approved):
👍

PR #1536, opened 2021-05-06
Run cypress on staging

trevorgerhardt:
Set up our Cypress tests to run against a remote URL (https://conveyal.dev) in addition to running locally. The new Github Action only runs on merges to master so that each PR isn't hammering a live instance and also because we don't always run a staging server during our development phase.

Currently, not all tests are passing, but enough are that it is useful enough to merge. The failing tests specific to the "Cypress Remote" tests will be addressed in subsequent pull requests.

The credentials currently used need to be manually updated on a monthly basis. It should only take a few minutes and the cycle will most likely correspond with our releases so it shouldn't be too hard to maintain for now.


trevorgerhardt:
#### 📝Changed routes:



- [**/api/auth/\[...auth0\]**](https://analysis-nuez1gpng-conveyal.vercel.app/api/auth/[...auth0])



Commit cd2ae2d6fa22540da598aec9891c6db5e6bc36ec (https://analysis-nuez1gpng-conveyal.vercel.app).

ansoncfit (approved):

PR #1541, opened 2021-05-10
Format analysis/modification errors

trevorgerhardt:
In the transition to the new Chakra version it appears we lost some formatting for returned "scenario modification errors" from an analysis request. This adds back the `disc` bullet points and formats the stack trace with `<pre>`.


ansoncfit (approved):
Errors (arising from analysis settings or scenario application) now show up consistently in the top of the analysis panel 👍

I may propose a few trivial message changes (e.g. "Errors were found in the project" → "Analysis errors") in a follow-up PR.

Issue #1282, opened 2020-10-07
Idea: Remove "auto-save" from modifications in favor of a manual save


trevorgerhardt:
Auto-save prevents "undoing" changes. If a user accidentally makes changes that they want to undo, theres no way to prevent the save. 

1. Keep the "Save on return to Modifications" action
2. Add an additional state indicating unsaved changes
3. Add buttons to "commit" or "discard" the changes

abyrd:
I do find that even though I know it auto-saves, every time I finish a modification I find myself looking for a way to save it, or at least confirm that it has been saved before I navigate away. Rather than a back arrow (which feels a bit like abandoning your changes) I think we should clearly label the button as "save and return to modifications". Another button could revert to last saved version. Very good point about being able to revert to a known state, it's easy to start working on something and realize you've put in wrong figures, and want to just go back to "the way things were".

ansoncfit:
As just discussed, this may help resolve database write issues.

trevorgerhardt:
Closed in #1572

PR #1548, opened 2021-05-13
Increase modification auto-save debounce from 10 to 30 seconds

trevorgerhardt:
It was noticed on slow connections that the debounce occurred frequently enough to disrupt a normal workflow. This triples the normal auto-save interval while maintaining the save functionality (auto-save on an interval and when leaving the page).


ansoncfit (approved):
Longer term: https://github.com/conveyal/analysis-ui/issues/1282

PR #1563, opened 2021-05-21
Set updatedAt / createdAt to match r5 models

trevorgerhardt:
`updatedAt` specifically was accidentally being overwritten with a `number` when updates occurred from the frontend.


abyrd:
Nice, good to have pinpointed where those errors were coming from.

PR #1567, opened 2021-06-02
Regional Page refactor + dependency upgrades

trevorgerhardt:
**NOTE BEFORE REVIEW** This turned into more than it was supposed to, but it should be purely a refactor and no change in functionality other than the aggregation area creation being removed (temporarily) as it is getting moved into spatial dataset upload. A review of all the 197 changed files is not worth your time! If you would like to test something, please create a new regional analysis and compare it.

## Regional Page Refactor
Moved into module structure that co-locates files by subject instead of type.  The main motivation was to clean up the regional page hierarchy that was always confusing me. So before:
```
/lib
  /components
    /analysis
      - legend.tsx
      - results-page.tsx
  - [all other components]
  /hooks
    - use-display-grid.ts
    - [all other hooks]
  /utils
    - predict-job-time-remaining.ts
    - [all other utils]
```

and after:
```
/lib
  /regional
    /components
      - legend.tsx
      - results-page.tsx
      - [regional only components]
    /hooks
      - use-display-grid.ts
      - [regional only hooks]
    /utils
      - predict-job-time-remaining.ts
      - [regional only utils]
```

There will always be top level `/lib/components` and `/lib/utils` folders for general, reusable components and utils that aren't only associated with one subject.

Essentially, changes to regional analysis related things in the future should mostly affect that `/lib/regional` folder. Should improve discoverability.

Additional changes here:
- Splitting API calls into `/queries` and `/mutations` folders (motivated by other JS front end projects doing the same).
- Asserting the query parameter associated with the selected analysis at the top level. (ex: does this selected regional analysis actually have the cutoff specified in the query parameter).
- Splitting the results page and the active page.

## Dependency Updates
Most libraries have been updated to their most recent version. A few notes: 
- The main exception is `react-leaflet`, which requires more of a refactor than I want to commit to right now.
- Prettier (the formatter) changed a few of its rules. I didn't realize how many files it would've effected, but you'll notice a lot of format only changes in those 197 files.
- The combination of the Jest update and Chakra update caused a few tests to fail for reasons not associated with the files themselves, instead with our set up. I replaced some of the snapshot tests with "Mount this component without error" tests, which I may move towards in the future more.


lib/aggregation-area/components/chart.tsx
```diff
@@ -123,9 +134,9 @@ export default function AggregateAccessibilityComponent({
       </FormControl>
 
       <AggregateAccessibilityReadout
-        name={regionalAnalysisName}
-        weightByName={weightByName}
-        accessToName={accessToName}
+        name={p.regionalAnalysisName}
+        weightByName={p.weightByName}
+        accessToName={p.accessToName}
```
ansoncfit, line 139: Padding:

image

ansoncfit (changes requested):
Overall regional analysis creation/comparison works on staging -- just a couple styling/padding issues I noticed.

lib/regional/components/active-analysis-page.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <AnalysisBounds analysis={analysis} />
+
+      <ActiveJob job={activeJob} />
+
+      <VariantSelectors
+        analysisVariant={analysisVariant}
+        onChangeCutoff={(v) => routeTo({cutoff: v})}
+        onChangePercentile={(v) => routeTo({percentile: v})}
+        onChangePointSet={(v) => routeTo({pointSetId: v})}
+      />
```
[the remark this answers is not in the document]
trevorgerhardt: Removed them while the job is running 👍

lib/aggregation-area/components/chart.tsx
trevorgerhardt: Fixed 👍

ansoncfit (approved):

PR #1576, opened 2021-06-24
When the Java API responds with a 401 or 403, log the user out

trevorgerhardt:
This PR forces a log out if the Java API responds with a 401 or 403 response. We were experiencing cases where the id token expired at a different rate than the client session. This should help mitigate user's having issues with this.


abyrd (approved):
Tested on staging, seems to be working well.

PR #1586, opened 2021-09-01
Repurpose Resources tab for DataSources

abyrd:
## Description

I have made a few changes on top of those @ansoncfit already made to provide minimum viable support for interacting with the DataSource endpoints on the backend, which are being developed in PR conveyal/r5#732. DataSources are basically just a new name for what were were originally calling Resources in the early design phase.

The data types on the backend allow for various types of DataSources, but currently we only create and manipulate SpatialDataSources, which correspond to vector GIS layers. The backend currently supports upload of ESRI Shapefile, GeoJSON, and GeoPackage to the DataSources endpoint.

Identifiers in the code have only been partially updated to reflect the Resources -> DataSources change. Specifically, API URLs and some method parameters have been changed to reflect the new name DataSources, and no references to SpatialSources should remain. But we do need another round of renaming of files etc.

## API endpoints

We plan to migrate some of these endpoints (especrially the GET actions) to the general purpose NextJS Mongo functions. The backend provides basic stopgap implementations of these endpoints. Mutating the datasource description should also be supported, but datasources are otherwise effectively immutable.

- `GET /api/datasource?regionId=[regionId]` fetches an array of all datasources for the given region ID
- `GET /api/datasource/[datasourceId]` fetches a single datasource
- `DELETE /api/datasource/[datasourceId]` deletes a single datasource. This must be done by the backend, not NextJS, because the backend has to actually delete files, not just a database record.
- `POST /api/datasource` create a new datasource; body is a form containing the fields `sourceName` (just a text label), `regionId` and `sourceFiles` (containing the uploaded files). 

While listing these out, I notice that almost every operation except "GET one datasource" needs a regionId, and the UI code situates datasource URLs hierarchically under regions. I discussed this pattern with @trevorgerhardt and he confirmed that the backend endpoints do not need to be changed to follow that pattern (regionId can be a query parameter where needed). 

The only operation we're currently supporting on datasources is converting polygons to aggregation areas (and this is only supported for shapefile uploads, not yet generalized). The endpoint to create an aggregation area is:

`POST /api/region/[regionId]/aggregationArea` with a body containing form fields `dataSourceId` (ID of a polygon datasource), `nameProperty` (the property of the datasource that contains the aggregation area names), and `zoom` (the zoom level). 

## Known issues

These changes still look very rough visually so the UI will need to be cleaned up. 

For some reason the UI always sends a "nameProperty=undefined" query parameter when calling `dispatch(toAggregationArea({resource, nameProperty, zoom}))`.


lib/db/authenticated-collection.ts
```diff
@@ -155,7 +161,7 @@ export default class AuthenticatedCollection {
   remove(_id: string) {
     return this.collection.deleteOne({
       accessGroup: this.accessGroup,
-      _id
```
ansoncfit, line 158: What is the impact of this change?

ansoncfit (changes requested):
Cypress tests started failing with 7efa3517. Running tests locally, it looks like the UI thinks bundles are deleted with the appropriate actions, but the corresponding bundles are not actually removed from the database. An issue with the multi-purpose remove nextjs lambda?

lib/db/authenticated-collection.ts
trevorgerhardt: Added a longer note about why this is necessary here. From our mixed use of `ObjectId`s and `string`s. One issue I'd like to mitigate ASAP...

abyrd:
Testing locally. When I upload a shapefile then convert it to aggregation areas, enabling "dissolve all features to a single area" the frontend sends a request with: `dataSourceId: 6148a47ae820fd5c503cce4f, nameProperty: the_geom, zoom: 9`. 

The request does not include the `mergePolygons=true` parameter (described at https://github.com/conveyal/r5/pull/732#issuecomment-915302846), so the backend does not merge the polygons and creates one area per polygon, named using the specified nameProperty field "the_geom" so it's calling toString on the geometry.

lib/dataSource/components/createAggregationArea.tsx
```diff
@@ -0,0 +1,154 @@
...
+  onClose
+}: {
+  dataSource: CL.SpatialDataSource
+  onClose: () => void
+}) {
+  const activity = useActivity()
+  const toAggregationArea = useConvertToAggregationArea(dataSource)
+  const [zoom, setZoom] = useState(SUPPORTED_ZOOM_LEVELS[0])
+  const [dissolveAll, setDissolveAll] = useState(true)
+  const defaultAttributeName = get(dataSource, 'attributes[0].name', '')
```
ansoncfit, line 65: To avoid having `the_geom` as the default property, should this include `.filter((a) => a.type == 'TEXT')`?

abyrd: We should consider removing the constraint that areas are named only from text fields. It seems reasonable to use a numeric ID as a polygon name, and the backend can handle this - it just converts any value to a string, as evidenced by what it does when the_geom is the name field. In fact the main constraint we may want is to disallow using `type:GEOM` attributes as names. Here we do need to prevent sending the request with a missing or GEOM nameProperty though.

abyrd:
I just pushed an update that will try to select the most appropriate name attribute. It looks for a text attribute called "name" (case insensitive), then falls back on the first text attribute if there are any. Numeric attributes are also allowed but must be selected by the user (the button is disabled until manual action is taken). I have tested this on shapefiles with and without text attributes, and with text attributes called "Name" with capital N. It seems to work as intended with latest r5#748 which is also patched to serve up the resulting mask grids at the URL expected by the UI.

trevorgerhardt (approved):

lib/constants/index.ts
```diff
@@ -32,16 +32,16 @@ export const API_URL = `${process.env.NEXT_PUBLIC_API_URL}/api`
 
 export const API = {
   Activity: `${API_URL}/activity`,
+  AggregationArea: `${API_URL}/aggregationArea`,
   Analysis: `${API_URL}/analysis`,
   Bundle: `${API_URL}/bundle`,
+  DataSource: `${API_URL}/dataSource`,
+  Files: `${API_URL}/files`,
```
ansoncfit, line 39: API.Files is no longer used, right?

PR #1590, opened 2021-09-13
Delete regional analyses from the Java API

trevorgerhardt:
A previous refactor to Regional Analyses changed it to use the Next.js API for deletion from MongoDB. This caused analyses to keep running because the job was not also stopped in the Broker. This PR changes it back to delete from the Java API.

## How to test

1. Create a regional analysis. While it is running, delete it. 
2. Ensure that the job is stopped from the Admin panel and that there are no errors.


abyrd (approved):
I started a backend instance and started a regional analysis. Once all workers had started and were reporting results I deleted the analysis via the UI. The job no longer appeared on the admin panel or API responses, and no errors were reported on the backend or in the Javascript console. Looks like it works.

PR #1591, opened 2021-09-16
Fix editing bounds in a region

trevorgerhardt:
Corrects the import from `lonlat` in `edit-bounds.js` and converts that file to TypeScript.

## How to test

1. Ensure you can drag the markers on the map to edit bounds in a region.

## Todo
- [ ] Add a Cypress test that drags the markers


ansoncfit (approved):
Successfully dragged bounds for region creation, region editing, and custom analysis area 👍

PR #1592, opened 2021-09-16
Remove `DEV` class from region bounds editor

trevorgerhardt:
This makes it no longer hidden. Thankfully it was hidden before cause my math was embarrassingly wrong 😅


ansoncfit (approved):
Area calculation looks sensible now 🗺 🔢 👍

PR #1594, opened 2021-09-17
Add description when origin points exceeds the maximum allowed

trevorgerhardt:
While editing a region bounds, show an error color and a description that the amount of origin points exceeds our maximum when they are over 5,000,000.

NOTE: Does not disable the form.


lib/components/edit-bounds-form.tsx
```diff
@@ -153,7 +155,7 @@ export default function EditBoundsForm(p: {
 }
 
 const originsToStatus = (originPoints: number): AlertStatus =>
-  originPoints >= 1_000_000
+  originPoints >= 5_000_000
```
ansoncfit, line 158:
```suggestion
  originPoints >= MAX_ALLOWED_ORIGIN_POINTS
```

ansoncfit (changes requested):

ansoncfit (approved):
👍

PR #1596, opened 2021-09-20
Set region ID on the client side

trevorgerhardt:
The region ID exists already on the client side, this sends it along with the analysis request so that it doesn't need to be looked up on the backend. Note: it was already specified for regional analysis creation.

Additionally, this PR updates the `analysis.properties` file to use the new config settings.


abyrd (approved):
Tested against staging, worker started up and was tagged on EC2 with all expected fields.

PR #1599, opened 2021-09-21
Add regional analysis migration

trevorgerhardt:
Convert old regional analyses that have single `cutoffMinutes`, `travelTimePercentile` and `grid` to have the array versions introduced in v5.10.0.


ansoncfit (approved):
For now, these are being run via Compass. We're saving the scripts for future reference.

PR #1603, opened 2021-09-24
Fix remove stops pattern selector

trevorgerhardt:
## How to test

1. Create a Remove Stops modification.
2. Add and remove Patterns to ensure it works as intended.


trevorgerhardt:
Tested on staging, merging myself 👍

PR #1607, opened 2021-09-28
Cache the connection Promise to help with serverless

trevorgerhardt:
Caching the connection `Promise` ensures that if multiple connections are received simultaneously, multiple clients don't get created while waiting for the first one to connect. 

This should not change functionality but adds comments and more closely aligns with the recommended best practices from Vercel and MongoDB


abyrd (approved):
I think in these matters we want to align as closely as possible with Vercel/Mongo advice. Adding comments and adhering to their suggested best practices sounds good to me. Approving so we can test this together with other recent changes.

PR #1608, opened 2021-09-28
Create new hook that queries each modification directly

trevorgerhardt:
When querying a list of modifications we will first make a query that only retrieves the list of modification IDs matching that query. After, we will fetch each modification by itself.


lib/modification/hooks/use-modifications.ts
```diff
@@ -0,0 +1,75 @@
...
+      if (res.ok === true) return res.data
+      else if (res.ok === false) throw res
+    })
+  )
+}
+
+/**
+ * Read only list of modifications. First queries via the filter for all IDs that match. Then it
+ * queries the modifications individually. A list of modifications could exceed the size limit of
+ * Vercel.
```
abyrd, line 29: This documentation might be confusing to read at a later date: "read only list" sounds like an immutable list (maybe that's actually what it means?), "a list of modifications" does not unambiguously refer to the modification contents instead of the IDs.
 
```suggestion
 * Fetch a list of modification IDs for modifications that match the filter, then fetch all the individual modifications
 * separately. An array containing the full text of many modifications has been  known to exceed Vercel's 
 * response body size limit.
```

abyrd (approved):
I have tested on the same project where I was encountering errors yesterday. I now see no HTTP response codes indicating errors. The request to `modifications` that was failing with a 500 due to excessive size now succeeds, as it contains only a list of IDs. This is immediately followed by fetching each modification separately by ID. Looks like an effective change. I was thinking this could even have some caching benefits on certain huge modifications, but upon returning to the modification list I see it always re-fetches every modification and there are no "unmodified" responses. Maybe this is due to cache settings elsewhere (even in my own dev tools).

I'm also not seeing the 400 errors on other endpoints that we were seeing yesterday - guessing those errors were cascading from the lambda handler failing on the `modifications` request that gave a 500.

I made one comment outside this review about some possible improvements to code comments.

PR #1610, opened 2021-09-28
Use scenarioJsonUrl endpoint to download Regional Analysis Request JSON from S3

trevorgerhardt:
Use new endpoint that retrieves scenario and modification JSON from S3 instead of the database.

## How to test

1. Navigate to a completed regional analysis.
2. Click "Download Results" and select "Scenario and modification JSON"

Related: https://github.com/conveyal/r5/issues/621


abyrd (approved):
Seems to be working well. Observations: Downloading from URLs in this way they are caught by popup blockers. Not just a warning about downloads, but the "new tab" is almost invisibly hidden until the user intervenes. I'm not sure if there's any way around that.

PR #1619, opened 2021-10-05
Modification "Save changes" refactor

trevorgerhardt:
Screen Shot 2021-10-06 at 15 23 02


abyrd:
Looks good, definitely usable as-is. I like the green save button with text and icon. I did notice that immediately after a save, when you make a change you get the text "Unsaved changes. Last saved less than a minute ago." which can wrap around to the next line, shifting the edit controls down. They move back up when the text changes to the shorter "Unsaved changes. Last saved 1 minute ago." This is just a minor layout detail, but it does lead me to think again: if there were always two rows present (1. Unsaved changes vs. No unsaved changes, 2. Last saved ___ ago) wrapping would probably be eliminated.

PR #1644, opened 2021-10-29
When downloading isochrones, return a feature collection with a point for the origin

### LINKED ISSUE #627: Allow users to download origin point with geojson isochrone
This would save some WRI users some time and effort.

trevorgerhardt:
Fixes #627


ansoncfit (approved):
When opening the downloaded file in QGIS, there is an option to select the point or isochrone layer 👍

image

PR #1647, opened 2021-11-01
Fix snap to existing stops

trevorgerhardt:
## How to test

1. Create an add trip pattern. Edit the geometry so that it passes over stops.
2. Click "Snap to existing stops". Ensure it snaps to stops that it passes over.


ansoncfit (approved):
Tested on staging 👍

